PR #23, opened 2023-05-18
Update column pattern regex to incorporate special column comments

drwl:
* Extracts `AnnotationDiffGenerator` out of `FileAnnotator`
* Updates the column pattern regex based on https://github.com/ctran/annotate_models/issues/946
* Adds some tests for `FileAnnotator`


drwl:
Note: working with column annotations are brittle due to the fact they're strings with varying white space. There's definitely work to be done to make this less brittle.

PR #29, opened 2023-05-23
Capture columns with multibyte comments

drwl:
Work in progress. Attempts to use logic from https://github.com/ctran/annotate_models/pull/986/.


drwl:
Stale so closing for now.

PR #40, opened 2023-06-10
Add `ANNOTATERB_SKIP_ON_DB_TASKS` ENV var to skip auto annotations

drwl:
In the old annotate gem, it used to be possible set `ANNOTATE_SKIP_ON_DB_MIGRATE` to `true|t|yes|y|1` to skip the automatic annotation that happens after a Rails db task is run (e.g. `bin/rails db:migrate`). Note that the automatic annotation would only happen if a user installed the rails plugin (done via `rails g annotate_rb:install`).

This behavior got removed when the dependency on ENV was removed. After some feedback, it seems like it's worth adding back to support special development setups or for checks in a CI pipeline.

Note: if you want the updated rake task, you'll have to rerun `rails g annotate_rb:install`.

### DIFF
--- a/dummyapp/lib/tasks/annotate_rb.rake
+++ b/dummyapp/lib/tasks/annotate_rb.rake
@@ -1,7 +1,7 @@
 # This rake task was added by annotate_rb gem.
 
 # Can set `ANNOTATERB_SKIP_ON_DB_TASKS` to be anything to skip this
-if Rails.env.development? && ENV['ANNOTATERB_SKIP_ON_DB_TASKS'].nil?
+if Rails.env.development? && ENV["ANNOTATERB_SKIP_ON_DB_TASKS"].nil?
   require "annotate_rb"
 
   AnnotateRb::Core.load_rake_tasks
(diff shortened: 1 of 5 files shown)


dummyapp/lib/tasks/annotate_rb.rake
```diff
@@ -0,0 +1,8 @@
+# This rake task was added by annotate_rb gem.
+
+# Can set `ANNOTATERB_SKIP_ON_DB_TASKS` to be anything to skip this
+if Rails.env.development? && ENV["ANNOTATERB_SKIP_ON_DB_TASKS"].nil?
```
Maimer, line 4: Minor: but I might have gone with:

```ruby
if Rails.env.development? && !ENV["ANNOTATERB_SKIP_ON_DB_TASKS"]
```

This is because you could set this environment variable to `false` and it would skip annotations, but probably not that big of a deal.

drwl: I don't know if it's shell/environment specific but I _think_ any values are interpreted as strings and not parsed.

```ruby
# $ A_STRING=false pry

[1] pry(main)> ENV['A_STRING']
=> "false"
```

Maimer: That's true, I guess that makes my comment not as relevant. (Also realized I had the boolean logic backwards and had included a `!`)

drwl: No worries. I wasn't sure how ENV behaved, so when I first read your comment I thought this could be changed as well.

PR #44, opened 2023-06-12
Prettify column defaults

### LINKED ISSUE #45: Column defaults change in migration
When migrating from the `annotate` gem to the `annotaterb` gem the default column values aren't represented in the same way as previously. My expectation would be that when migrating, as long as I have the same configuration setup, that the results would be the same as the previous gem. The output from the `annotate` gem "humanized" the defaults instead of directly representing what is the in database. For example instead of showing a default of `0` or `1` for a boolean field, as that is stored as a `TINYINT`, it would display it as `TRUE` or `FALSE`. Here are some example diffs from columns with default fields:

Boolean column:

```diff
-#  enabled              :boolean          default(TRUE), not null
+#  enabled              :boolean          default("1"), not null
```  

Integer column:

```diff
-#  lock_version         :integer          default(0), not null
+#  lock_version         :integer          default("0"), not null
```

Enum column:

```diff
-#  initiation_type      :integer          default("one_time"), not null
+#  initiation_type      :integer          default("0"), not null
```

## Commands

```
$ bundle exec annotate models
```

## Version

- `annotaterb` version: `4.3.0`
- `rails` version: `7.0.4.3`
- `ruby` version: `3.1.3`

drwl:
Fixes #45. 

When I extracted out column annotation building [1], I changed the way the default value was read. `Model#column_defaults` has the column default values that are casted appropriately, where as `Column#default` is uncasted and straight from the database, most of the time as a String. It's important to note that the various adapters (sqlite, mysql, postgres) implement the columns differently [2], especially as it relates to Column#default and Model#column_default values.

This PR restores parity with the original gem on how defaults are handled. It also adds tests for the various adapters, because the implementations vary under the hood.

As an example, the mysql adapter sets the value for an integer field without a default to be `0` in `Model#column_defaults`, where as sqlite and postgres set it to be `nil`.

[1] https://github.com/drwl/annotaterb/pull/28/files#diff-92c15977800c36cb911c1d7971ede0da49dae93b0c9e555777a5ca67ee81d93dR12-R14

[2] https://gist.github.com/drwl/ef274a763b773d132abfdb95d32347d2

### DIFF
--- a/dummyapp/Gemfile.lock
+++ b/dummyapp/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
-    annotaterb (4.2.0)
+    annotaterb (4.3.0)
 
 GEM
   remote: https://rubygems.org/
--- a/dummyapp/app/models/case.rb
+++ b/dummyapp/app/models/case.rb
@@ -1,0 +1,19 @@
+# == Schema Information
+#
+# Table name: cases
+#
+#  id               :integer          not null, primary key
+#  default_number   :integer          default(1)
+#  default_zero     :integer          default(0)
+#  is_default_false :boolean          default(FALSE)
+#  is_default_true  :boolean          default(TRUE)
+#  json_text_field  :text
+#  simple_bool      :boolean
+#  simple_int       :integer
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
+#
+class Case < ApplicationRecord
+  # Model file with fields to test annotations
+  serialize :json_text_field, JSON, default: {}
+end
--- a/dummyapp/app/models/task.rb
+++ b/dummyapp/app/models/task.rb
@@ -1,2 +1,13 @@
+# == Schema Information
+#
+# Table name: tasks
+#
+#  id         :integer          not null, primary key
+#  content    :string
+#  count      :integer
+#  status     :boolean
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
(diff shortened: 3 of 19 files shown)


drwl:
Some notes from debugging this. The default value as a String gets passed from ActiveRecord internally itself.

```ruby
# The column object we get from AR has a default value that's String
# [4] pry(#<AnnotateRb::ModelAnnotator::ColumnAnnotation::AttributesBuilder>)> rcol = @column.instance_variable_get("@column")
# => #<ActiveRecord::ConnectionAdapters::PostgreSQL::Column:0x000000010e19ea60
#  @collation=nil,
#  @comment=nil,
#  @default="0",
#  @default_function=nil,
#  @generated="",
#  @name="duration_minutes",
#  @null=true,
#  @serial=nil,
#  @sql_type_metadata=
#   #<ActiveRecord::ConnectionAdapters::SqlTypeMetadata:0x000000010e19f140
#    @limit=4,
#    @precision=nil,
#    @scale=nil,
#    @sql_type="integer",
#    @type=:integer>>

# Internally, it also is a String
# [4] pry(#<ActiveRecord::ConnectionAdapters::PostgreSQLAdapter>)> field
# => ["duration_minutes", "integer", "0", false, 23, -1, nil, nil, ""]
```

drwl:
As a side note, I found out that column names should actually be strings. There are some tests that use `mock_column(:some_symbol, ...)` and this leads to incorrect test setup + expected output.

Will handle in another PR.

PR #50, opened 2023-06-21
Add support for table comments

drwl:
This PR makes the following changes:
* Adds support for table comments
* Adds 2 new options keys: `with_column_comments` and `with_table_comments`
* Checks for `with_comment: true, with_column_comments: true` to add column comments
* Checks for `with_comment: true, with_table_comments: true` to add table comments
* Changes `Options#load_defaults`, it defaults `with_column_comments` and `with_table_comments` to use the value set in `with_comment` if not previously set

### DIFF
--- a/dummyapp/Gemfile.lock
+++ b/dummyapp/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
-    annotaterb (4.3.0)
+    annotaterb (4.3.1)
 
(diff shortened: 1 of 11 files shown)


drwl:
Credit goes to @suzunedev from https://github.com/ctran/annotate_models/pull/987 for PR inspiration. Addresses https://github.com/ctran/annotate_models/issues/917.

Issue #57, opened 2023-09-01
Default array value is double-quoted/escaped


manuelmeurer:
When I set the default value for an array text column via the Attributes API, the annotate output is double-quoted (or quoted and escaped?).

```ruby
attribute :notifications, default: %w(newsletter)
```

generates

```
#  notifications :text             default(["\"newsletter\""]), not null, is an Array
```

## Commands

```
$ bundle exec annotaterb models
```

## Version

- annotaterb version: 4.4.0
- rails version: 6.1.7.6
- ruby version: 3.2.2

drwl:
Hi @manuelmeurer, thanks for taking the time to report an issue! I'll take a further look into this soon. For reference, what database/adapter are you using?

manuelmeurer:
Postgres 15. Thanks for taking a look!

It has something to do with https://github.com/drwl/annotaterb/blob/dc95dfff886adeebea5a1a551ceaaf0a9dc2c794/lib/annotate_rb/model_annotator/column_annotation/column_wrapper.rb#L98 but I couldn't figure it out...

drwl:
Could you share what the line used to described that field in `schema.rb`?

This is what I'm using while debugging and wanted to double check it matches:
```ruby
# schema.rb
create_table "test_cases", force: :cascade do |t|
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.string "default_string_array", default: ["something"], array: true
  end
```

```ruby
# test_case.rb
class TestCase < ApplicationRecord
  attribute :default_string_array, default: %w(newsletter)
end
```

manuelmeurer:
The default is set to `[]` in the migration and schema (I'm not a fan of setting defaults in the DB), only via the Attributes API to `%w(newsletter)`

```ruby
# migration
create_table :mytable do |t|
  t.text :notifications, array: true, default: [], null: false
end

# schema
create_table "mytable", force: :cascade do |t|
  t.text "notifications", default: [], null: false, array: true
end

# model
attribute :notifications, default: %w(newsletter)
```

Another questions is whether annotate should show the default as `["newsletter"]` in this case at all... now that I think about it, I think it should be `[]` since that is the migration/schema default.

drwl:
Not sure if you saw the tag in the PR but if you have the time and are able to test the branch, I'd be happy to merge it in after.

[...]

PR #58, opened 2023-09-02
Fix the default array value from being escaped

drwl:
The Postgres adapter supports `array: true` for various column types [1]. In #57, it was shown that annotaterb adds escaped string values. 

[1] https://guides.rubyonrails.org/active_record_postgresql.html#array

### DIFF
--- a/lib/annotate_rb/model_annotator/column_annotation.rb
+++ b/lib/annotate_rb/model_annotator/column_annotation.rb
@@ -7,6 +7,7 @@ module AnnotateRb
       autoload :TypeBuilder, "annotate_rb/model_annotator/column_annotation/type_builder"
       autoload :ColumnWrapper, "annotate_rb/model_annotator/column_annotation/column_wrapper"
       autoload :AnnotationBuilder, "annotate_rb/model_annotator/column_annotation/annotation_builder"
+      autoload :QuotedDefaultValueBuilder, "annotate_rb/model_annotator/column_annotation/quoted_default_value_builder"
     end
   end
 end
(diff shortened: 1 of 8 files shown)


drwl:
Notes that I took while debugging:

```ruby
# I have a model with a column that looks like this:
# In schema.rb
create_table "test_cases", force: :cascade do |t|
  t.string "default_string_array", default: ["something"], array: true
end

# In the model file
class TestCase < ApplicationRecord
  attribute :default_string_array, default: %w(newsletter)
end

# [12] pry(main)> col = columns.last
# => #<ActiveRecord::ConnectionAdapters::PostgreSQL::Column:0x000000010803d028
#   @collation=nil,
#     @comment=nil,
#     @default="{something}", <= when schema.rb has default: [], @default is "{}"}
#     @default_function=nil,
#     @generated="",
#     @name="default_string_array",
#     @null=true,
#     @serial=nil,
#     @sql_type_metadata=
#       #<ActiveRecord::ConnectionAdapters::SqlTypeMetadata:0x000000010803d708
#       @limit=nil,
#     @precision=nil,
#     @scale=nil,
#     @sql_type="character varying[]",
#     @type=:string>>

# What's interesting is the default is the one in schema.rb, but if we look at `TestCase.column_defaults`,
#   it shows "newsletter"

TestCase.column_defaults
# => {"id"=>nil, "created_at"=>nil, "updated_at"=>nil, "default_string_array"=>["newsletter"]}
```

drwl:
What was interesting to me is that the postgres adapter returns the column default value as `"{}"`, although the "correct" default is still returned in `ModelName.column_defaults`.

drwl:
@manuelmeurer by chance would you be able to run annotaterb referencing this branch in your Gemfile to see if it annotates correctly? It works for me locally but would like an extra check before merging.

manuelmeurer:
It fixes the problem for array columns but now string defaults don't show the quotes anymore.
zsh 2023-09-05 22-29-33

drwl:
Thank you for testing 👍, I'm surprised there were no unit tests that caught that. I'll look into adding those.

drwl:
@manuelmeurer could you test the latest version of this branch? I think it should fix it + not have regressions.

manuelmeurer:
@drwl Yes, that fixes it, good job! 😄

PR #65, opened 2023-09-14
Remove `ARGV` use during runtime

drwl:
This PR removes `ARGV` from being used implicitly during runtime. Before this change, ARGV would be used and modified in `Parser#parse` which made it hard to reason about how `ModelFilesGetter#list_model_files_from_argument` works.

`ARGV` is still used at the time of execution but gets cloned in `Parser#new`.

### DIFF
--- a/lib/annotate_rb/model_annotator/model_files_getter.rb
+++ b/lib/annotate_rb/model_annotator/model_files_getter.rb
@@ -41,9 +41,9 @@ module AnnotateRb
         private
 
         def list_model_files_from_argument(options)
-          return [] if options[:working_args].empty?
+          return [] if options.get_state(:working_args).empty?
 
-          specified_files = options[:working_args].map { |file| File.expand_path(file) }
+          specified_files = options.get_state(:working_args).map { |file| File.expand_path(file) }
 
           model_files = options[:model_dir].flat_map do |dir|
             absolute_dir_path = File.expand_path(dir)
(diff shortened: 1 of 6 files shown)


lib/annotate_rb/parser.rb
```diff
@@ -35,11 +35,11 @@ def self.parse(args, existing_options = {})
     }.freeze
 
     def initialize(args, existing_options)
-      @args = args
+      @args = args.clone
```
drwl, line 38: Before this change, `@args` was a reference to `ARGV`. This resulted in `parse_command` and `parser.parse!` to modify `ARGV`.

lib/annotate_rb/runner.rb
```diff
@@ -9,14 +9,15 @@ def run(args)
...
-      parsed_options = Parser.parse(args)
+      parser = Parser.new(args, {})
+
+      parsed_options = parser.parse
+      remaining_args = parser.remaining_args
 
       options = config_file_options.merge(parsed_options)
 
-      @options = Options.from(options, {})
+      @options = Options.from(options, {working_args: remaining_args})
```
drwl, line 20: Could probably come up with a better name

lib/annotate_rb/model_annotator/model_files_getter.rb
```diff
@@ -9,13 +9,9 @@ class << self
         # of model files from root dir. Otherwise we take all the model files
         # in the model_dir directory.
         def call(options)
-          model_files = []
+          model_files = list_model_files_from_argument(options)
 
-          # Note: This is currently broken as we don't set `is_rake` anywhere.
-          # It's an artifact from the old Annotate gem and how it did control flow.
-          model_files = list_model_files_from_argument(options) if !options[:is_rake]
```
drwl, line 16: `:is_rake` is a legacy option from the old Annotate gem, it's not used in this gem.

PR #69, opened 2023-10-13
Add integration test for rake task installer

drwl:
This PR adds integration test for the rake task installer. The installer is invoked using this CLI command:

`bin/rails generator annotate_rb:install`

### DIFF
--- a/spec/integration/rails_generator_install_spec.rb
+++ b/spec/integration/rails_generator_install_spec.rb
@@ -31,4 +31,22 @@ RSpec.describe "Generator installs rake file", type: "aruba" do
     expect(last_command_started).to be_successfully_executed
     expect(installed_rake_task).to eq(actual_rake_task)
   end
+
+  context "when the rake task already exists" do
+    before do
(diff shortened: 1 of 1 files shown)


spec/integration/rails_generator_install_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    before do
+      touch(rake_task_file)
+    end
+
+    it "returns the Thor cli" do
+      # First check that the file exists in dummyapp
+      expect(exist?(rake_task_file)).to be_truthy
+
+      # TODO: Improve this so we don't have to rely on `exit_timeout`
+      _cmd = run_command(generator_install_command, exit_timeout: 3)
```
drwl, line 45: When the rake task already exists and the generator command is ran by Aruba, Thor (which Rails uses) provides an interactive menu to the user on how to proceed.

Currently, I don't have the knowledge or expertise to proceed with code flow without the use of `exit_timeout`. There's probably a better way to do this.

PR #72, opened 2023-10-25
Fix annotations swallowing comments

drwl:
This PR attempts to fix a long standing bug that was carried over from the original gem. Comments near an annotation would get removed when adding or updating annotations.

https://github.com/ctran/annotate_models/issues/951, https://github.com/ctran/annotate_models/#warning

**Context**
Prior to this change, the AnnotateRb and the old gem Annotate would annotate model files in 1 of 2 ways:
1. The "update" method, existing annotations would be found using a regex and then new annotations would be changed in place. If a human made comment was added after the annotation block **without** a new line, it would get removed during the update process.
2. The "overwrite" method, which would happen when annotations did not exist OR if the `:force` option was set to true. A regex would be used to find annotation, remove it from the file, and then add annotations again to the file depending on options such as `:position`.

Relying on regexes to identify annotations in a file made it hard to reason about and hard to change, so most of them were removed and replaced with a custom Ruby file parser using Ripper [1], which makes working with Ruby files deterministic.

This PR makes the following behavioral changes:
- When updating annotations (i.e. annotations exist && `force: false`), it only updates the annotations and should no longer delete any human comments that are before or after an annotation block
- When generating (no annotations exist in the file) or re-generating annotations (`force: true`), it removes the annotation block and surrounding whitespace, if it exists, and then depending on the `position` option, will add annotations before the class declaration or after the class declaration. In the case of re-generating annotations, any surrounding whitespace will be preserved assuming annotations are being written to the same position.

[1] A great write up on Ripper: https://kddnewton.com/2022/02/14/formatting-ruby-part-1.html


drwl:
This PR is quite big and a good amount of the lines added are due to test cases I added to document the behavior when annotating a single file. There's a bit of polish left to do in terms of refactors, but I will add that in a future PR.

PR #88, opened 2024-02-18
Support non-model files in `CustomParser`

drwl:
This PR adds support for other Ruby files types that should be able to get model annotations added to them (i.e. factories, fabricators, etc).

In #72, AnnotateRb was changed to use `FileParser::CustomParser` to parse Ruby files instead of relying on regexes. That PR only added support for model files that used either `module Namespace...` or `class ModelName...` as the first line of Ruby. The parser did not have support for other file structures like FactoryBot factories or Fabrication fabricators. 

There may be other Ruby files that have code that is not currently handled by `CustomParser`, so those will have to be added in the future.

### DIFF
--- a/lib/annotate_rb/model_annotator/file_parser/custom_parser.rb
+++ b/lib/annotate_rb/model_annotator/file_parser/custom_parser.rb
@@ -138,13 +138,8 @@ module AnnotateRb
         # ```
         def on_do_block(block_var, bodystmt)
           if block_var.blank? && bodystmt.blank?
-            _const, last_lineno = @block_ends.last
-
-            # Minor optimization, only add "end" when a closing block isn't added by #on_method_add_block
-            if last_lineno != lineno
-              @block_ends << ["end", lineno]
-              add_event(__method__, "end", lineno)
-            end
+            @block_ends << ["end", lineno]
+            add_event(__method__, "end", lineno)
           end
           super
         end
--- a/lib/annotate_rb/model_annotator/file_parser/custom_parser.rb
+++ b/lib/annotate_rb/model_annotator/file_parser/custom_parser.rb
@@ -11,7 +11,7 @@ module AnnotateRb
 
         class << self
           def parse(string)
-            _parser = new(string).tap(&:parse)
+            _parser = new(string, "", 0).tap(&:parse)
           end
         end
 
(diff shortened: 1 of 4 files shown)


drwl:
### Adding support for other Ruby file structures
Load the file content into a string
```ruby
require 'ripper' # Want to use this to debug

content = File.readlines('file_name').join

Ripper.sexp(content)
# => Might look something like this
# => [:program,
#     [[:method_add_block,
#       [:command,
#        [:@ident, "factory", [1, 0]],
#        [:args_add_block,
#         [[:symbol_literal, [:symbol, [:@ident, "user", [1, 9]]]],
#          [:bare_assoc_hash,
#           [[:assoc_new,
#             [:@label, "aliases:", [1, 15]],
#             [:array,
#              [[:symbol_literal, [:symbol, [:@ident, "author", [1, 26]]]],
#               [:symbol_literal, [:symbol, [:@ident, "commenter", [1, 35]]]]]]]]]],
#         false]],
```

Match the left most symbol with an event [1] and add the `on_<event_name>` method in `CustomParser`. 

To debug it works, you can introspect the various instance variables which should have information that's used and additionally some code used for debugging.
```ruby
pry(main)> Parser.parse(content)
=> #<AnnotateRb::ModelAnnotator::FileParser::CustomParser:0x0000000158a40110
 @_const_event_map=
  {0=>[["factory", :on_command]],
   1=>[["first_name", :on_method_add_arg], ["first_name", :on_method_add_block]],
   2=>[["last_name", :on_method_add_arg], ["last_name", :on_method_add_block]],
   3=>[["date_of_birth", :on_method_add_arg], ["date_of_birth", :on_method_add_block]],
   4=>[["end", :on_do_block], ["factory", :on_method_add_block]]},
 @_input=
  "factory :user, aliases: [:author, :commenter] do\n  first_name { \"John\" }\n  last_name { \"Doe\" }\n  date_of_birth { 18.years.ago }\nend\n",
 @_stack_code_block=[],
 @block_ends=[["first_name", 1], ["last_name", 2], ["date_of_birth", 3], ["end", 4]],
 @block_starts=[["factory", 0], ["first_name", 1], ["last_name", 2], ["date_of_birth", 3], ["factory", 4]],
 @comments=[],
 @const_type_map={}>
```

Additionally, one can add a binding statement inside `def on_program` as that method gets called after a file finishes parsing.

[1] https://kddnewton.com/ripper-docs/events

Issue #97, opened 2024-03-20
Feature request: better custom type representation


viralpraxis:
Describe your problem here.

## Commands

We've been using annotate for some time alongside with custom AR types (via `ActiveRecord::Type.register`). Unfortunately, annotate was serializing values of these types into extremely long strings:

```
# currency       :string           default(#<Money::Currency id: usd, priority: 100, symbol_first: false, thousands_separator: ., html_entity: &#x20BD;, decimal_mark: ,, name: United States Dollar, symbol: $, subunit_to_unit: 100, exponent: 2, iso_code: RUB, iso_numeric: 643, subunit: Cent, smallest_denomination: 1, format: %n %u>), not null
```

We workaround the issue by monkey-patching ``AnnotateModels.quote`:

```ruby
      ....
      when BigDecimal               then value.to_s("F")
      when Array                    then value.map { |v| quote(v) }
      when Locale                   then value.to_s.inspect # custom types start here
      when Region                   then value.to_s.inspect
      when Billing::Plan            then value.to_s.inspect
```

so annotations were good to look at:

```
#  currency       :string           default("USD"), not null
```

We are looking forward migration to annotaterb so we'd also want to get rid of that monkey-patch. Perhaps an addtional configuration option would be sufficient?

## Version

- annotaterb version: 4.6.0
- rails version: 7.1.3.2
- ruby version: 3.3.0

drwl:
Hi @viralpraxis, thanks for checking out the gem and adding a featuring request!

Maybe we can brainstorm a solution together. I think adding a configuration option that takes a list of classes/types that gets mapped to `value.to_s.inspect` could work. I'm curious if there may be other custom mappings that might be desired as well.

On the other hand, adding some way to load some monkeypatch could bridge the gap for folks looking to migrate from the old annotate gem. So I can see value there.

Thoughts?

viralpraxis:
> I think adding a configuration option that takes a list of classes/types that gets mapped to value.to_s.inspect could work

I'm pretty sure this approach would do the trick. One thing I'm not sure about is how specific the configuration option should be. It might be as simple as

```yaml
classes_with_to_s_representation:
  - Locale
  - Region
```

or it might be more flexible:
```yaml
custom_type_default_value_representation:
  - class: "Locale"
    methods: ["to_s", "inspect"] # value.to_s.inspect
  - class: "Region"
    methods: "to_s" # value.to_s
```

[...]

PR #98, opened 2024-03-26
Add configurable classes list with `to_s` representation

viralpraxis:
Will resolve https://github.com/drwl/annotaterb/issues/97

### DIFF
--- a/lib/annotate_rb/options.rb
+++ b/lib/annotate_rb/options.rb
@@ -136,7 +136,8 @@ module AnnotateRb
       :target_action,
       :wrapper,
       :wrapper_close,
-      :wrapper_open
+      :wrapper_open,
+      :classes_default_to_s
     ].freeze
 
     PATH_OPTION_KEYS = [
--- a/lib/annotate_rb/options.rb
+++ b/lib/annotate_rb/options.rb
@@ -76,7 +76,8 @@ module AnnotateRb
       target_action: :do_annotations, # Core; Possible values: :do_annotations, :remove_annotations
       wrapper: nil, # ModelAnnotator, RouteAnnotator
       wrapper_close: nil, # ModelAnnotator, RouteAnnotator
-      wrapper_open: nil # ModelAnnotator, RouteAnnotator
+      wrapper_open: nil, # ModelAnnotator, RouteAnnotator,
+      classes_default_to_s: [] # ModelAnnotator
     }.freeze
 
     PATH_OPTIONS = {
--- a/lib/annotate_rb/parser.rb
+++ b/lib/annotate_rb/parser.rb
@@ -127,15 +127,6 @@ module AnnotateRb
         @options[:wrapper_open] = wrapper_open
       end
 
-      option_parser.on("--classes-default-to-s class",
-        "Custom classes to be represented with `to_s`, may be used multiple times") do |klass|
-        @options[:classes_default_to_s] = if @options[:classes_default_to_s].present?
-          [@options[:classes_default_to_s], ::Object.const_get(klass)]
-        else
-          [Object.const_get(klass)]
-        end
-      end
-
       option_parser.on("--wc",
         "--wrapper-close STR",
         "Annotation wrapper closing") do |wrapper_close|
@@ -256,6 +247,15 @@ module AnnotateRb
       option_parser.on("--without-table-comments",
         "exclude table comments in model annotations") do
         @options[:with_table_comments] = false
+      end
+
+      option_parser.on("--classes-default-to-s class",
+        "Custom classes to be represented with `to_s`, may be used multiple times") do |klass|
+        @options[:classes_default_to_s] = if @options[:classes_default_to_s].present?
+          [*@options[:classes_default_to_s], klass]
+        else
+          [klass]
+        end
       end
     end
 
--- a/lib/annotate_rb/parser.rb
+++ b/lib/annotate_rb/parser.rb
@@ -9,7 +9,7 @@ module AnnotateRb
 
     BANNER_STRING = <<~BANNER.freeze
       Usage: annotaterb [command] [options]
-      
+
       Commands:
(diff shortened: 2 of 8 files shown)


drwl:
Thanks for adding these. I think the changes look good to me, would you be able to add unit tests as well?

viralpraxis:
Sure. I opened this draft only to get high-level design approval, gonna fix CI and some tests soo

viralpraxis:
@drwl

drwl:
@viralpraxis Thanks for adding tests. I've had a chance to test these changes, I'll add some feedback in PR as well. I ran into some issues around getting constantizing class strings that I think we should address.

(I added a `binding.pry` to `lib/annotate_rb/runner.rb` on L20 before the bootstrap call)

How I tested, I ran through CLI passing in different classes that exist:
`DATABASE_ADAPTER=sqlite3 be annotaterb models --classes-default-to-s Integer --classes-default-to-s String`
This works great, I'll add another note in the PR, but we may want to append when it already exists.

When running with a class/type that isn't already in the load path then it will fail:
`DATABASE_ADAPTER=sqlite3 be annotaterb models --classes-default-to-s Integer --classes-default-to-s TStruct`
TStruct is a dummy model class in `app/models`. This results in NameError because it can't resolve the constant. `uninitialized constant TStruct (NameError)`

When loading by using config file, the strings don't get constantized:
```yml
# .annotaterb.yml
...
:classes_default_to_s: [Integer, String]
...
```

```ruby
 [1] pry(#<AnnotateRb::Runner>)> @options
 => #<AnnotateRb::Options:0x000000015a01d550
 @options=
   ...
   :classes_default_to_s=>["Integer", "String"],
```

I think at the core, we want to solve the issue of when to constantize types into their Class names. Perhaps keep things as strings and constantize inside `DefaultValueBuilder`? I don't have a strong opinion about this yet.

lib/annotate_rb/options.rb
```diff
@@ -76,7 +76,8 @@ def from(options = {}, state = {})
       target_action: :do_annotations, # Core; Possible values: :do_annotations, :remove_annotations
       wrapper: nil, # ModelAnnotator, RouteAnnotator
       wrapper_close: nil, # ModelAnnotator, RouteAnnotator
-      wrapper_open: nil # ModelAnnotator, RouteAnnotator
+      wrapper_open: nil, # ModelAnnotator, RouteAnnotator,
+      classes_default_to_s: [] # ModelAnnotator
     }.freeze
```
drwl, line 82: We'll also want to add the configuration key name to `OTHER_OPTION_KEYS` so that it gets picked up in default options.

lib/annotate_rb/parser.rb
```diff
@@ -127,6 +127,15 @@ def add_wrapper_options_to_parser(option_parser)
         @options[:wrapper_open] = wrapper_open
       end
 
+      option_parser.on("--classes-default-to-s class",
+        "Custom classes to be represented with `to_s`, may be used multiple times") do |klass|
+        @options[:classes_default_to_s] = if @options[:classes_default_to_s].present?
+          [@options[:classes_default_to_s], ::Object.const_get(klass)]
```
drwl, line 133: When there's multiple instances of this CLI option it will keep wrapping the previous value:

`DATABASE_ADAPTER=sqlite3 be annotaterb models --classes-default-to-s Integer --classes-default-to-s String`

```ruby
[1] pry(#<AnnotateRb::Runner>)> @options
=> #<AnnotateRb::Options:0x0000000143f19db8
 @options=
  {:target_action=>:do_annotations,
   :exit=>false,
   :original_args=>["models", "--classes-default-to-s", "Integer", "--classes-default-to-s", "String"],
   :classes_default_to_s=>[[Integer], String],
```

Notice that `Integer` is wrapped in an array. We may want to append here.

lib/annotate_rb/parser.rb
```diff
@@ -127,6 +127,15 @@ def add_wrapper_options_to_parser(option_parser)
         @options[:wrapper_open] = wrapper_open
       end
 
+      option_parser.on("--classes-default-to-s class",
```
drwl, line 130: Could we move this block outside of `#add_wrapper_options_to_parser`? Perhaps into `#add_model_options_to_parser` instead?

drwl (changes requested):

lib/annotate_rb/parser.rb
viralpraxis: Thanks, missed it completely. I used splat operator (*) because I prefer to avoid mutations if possible, let me note if you think appending is better in this case

lib/annotate_rb/parser.rb
viralpraxis: Done 👍🏻

lib/annotate_rb/options.rb
viralpraxis: Done

viralpraxis:
> I think at the core, we want to solve the issue of when to constantize types into their Class names. Perhaps keep things as strings and constantize inside DefaultValueBuilder? I don't have a strong opinion about this yet.

I tried to remove ::Object.get_const completely by comparing `value.class.name` with provided array of classes (without constantizing them). Will it work?

lib/annotate_rb/parser.rb
drwl: Splat operator also works 👍

drwl (approved):

drwl:
> > I think at the core, we want to solve the issue of when to constantize types into their Class names. Perhaps keep things as strings and constantize inside DefaultValueBuilder? I don't have a strong opinion about this yet.
> 
> I tried to remove ::Object.get_const completely by comparing `value.class.name` with provided array of class names (without constantizing them). Will it work?

I think this approach works great 👍 . Thanks for reporting the issue and putting up PR. I'll cut a gem release soon.

PR #107, opened 2024-04-12
Make `exclude_tests` option able to override other exclude options

drwl:
This PR adds support for `exclude_tests` option to be a boolean or an array. When it's a boolean, the behavior should stay the same. When it's an Array, any of the following options can be included:

Prior to this change, `exclude_scaffolds` would used for controller tests, request specs, and routing specs. With this change, it will depend on `exclude_scaffolds` AND `exclude_tests` assuming `exclude_tests` is an array.

```ruby
:model, :controller, :serializer, :request, :routing
```

Because it's an exclusion list, options being present will mean to exclude them. Where as the absence of an option means to include it. Options specified or not specified here will take precedence over their original option; more explained below.

_I think this is very confusing so changing the `exclude_` options to become `include_` should be something to consider for the future_

To illustrate the behavior:
When `exclude_tests: [:model]` => this means to include controller tests, serializer tests, request specs, and routing specs. This will take precedence over `exclude_scaffolds: true`.

When `exclude_tests: [serializer] and exclude_serializers: false`, this means that serializers will be included but their tests will be excluded.

**Note: when setting exclude_tests to an Array, to use the singular form and not the plural form.**

For example, to exclude controller tests:
```yaml
:exclude_tests: [controller]
```


JordanFinch72:
@drwl I'm not able to test this locally, but the described changes sound logical to me! This will definitely help to more granularly determine what files get annotated. It would definitely be a big plus to invert the exclusion to an inclusion (in my opinion), but that would definitely warrant a bump in the major version so as not to disrupt people's existing configuration.

On that note, has this been tested with a range of configurations in the old version? I'm paranoid that this code change might break people's current configurations with an upgrade to the gem; but if it's backwards-compatible, then brilliant!

drwl:
@JordanFinch72 I can do a quick check and test with some basic configurations but I believe it /should/ work. In the worst case, it can be part of a gem revision that gets reverted. I'll aim to get this in soon after testing again. I appreciate your feedback.

JordanFinch72:
It might be neat to have a spec for every combination of configurations, using matrices like:

| | model | controller | serializer |
|--------|--------|--------|--------|
| self | include | include | include |
| tests | include | include | include |

| | model | controller | serializer |
|--------|--------|--------|--------|
| self | include | include | include |
| tests | exclude | exclude | exclude |

| | model | controller | serializer |
|--------|--------|--------|--------|
| self | exclude | exclude | exclude |
| tests | include | include | include |

| | model | controller | serializer |
|--------|--------|--------|--------|
| self | exclude | exclude | exclude |
| tests | exclude | exclude | exclude |

Some of these would fail with the current `main`, but would presumably pass with this implementation and the correct configurations.

drwl:
@JordanFinch72 apologies for the delay, I've been focusing on job interviews + prep. Will answer your questions inline:

> I'm not able to test this locally

Cann you elaborate what you mean by this?

> On that note, has this been tested with a range of configurations in the old version? 

I believe so, I added new test cases to test that the behavior is backwards compatible

> It might be neat to have a spec for every combination of configurations, using matrices like:

This approach definitely makes sense, but I intend to add change the configuration to be inclusion based so that it's easier to reason about and use. I wanted to get this change in to unblock your current usage. So I don't currently plan on investing too much time into this (given the planned rewrite ahead).

JordanFinch72:
> @JordanFinch72 apologies for the delay, I've been focusing on job interviews + prep. Will answer your questions inline

Good luck! You can add "I deal with random developers and their pesky problems" to your CV!

> > I'm not able to test this locally
> 
> Cann you elaborate what you mean by this?

Sure, that was misleading. I only really have projects for work on my work laptop and I can't really clone your changes until they're released as a new version of the gen because of... reasons. I do apologise, as it's rather poor form of me to ask for changes and then to not be able to test them out.

>So I don't currently plan on investing too much time into this (given the planned rewrite ahead).

Fair enough, sounds reasonable!

drwl:
> ... they're released as a new version of the gem because of... reasons. I do apologise, as it's rather poor form of me to ask for changes and then to not be able to test them out.

No worries -- I'm happy to work with you on this. Just to be clear is it only officially published gems to a gem server like Rubygems? If not, I wonder if it's possible for you to pull based a branch: https://stackoverflow.com/questions/2823492/install-gem-from-github-branch (in case you're not familiar with this yet)

JordanFinch72:
Cool! I didn't know you could do that, thanks for the knowledge!

Just to check that I'm not being completely tired: is this the new format:


```
...
:exclude_fixtures: false
:exclude_helpers: true
:exclude_scaffolds: true
:exclude_serializers: false
:exclude_sti_subclasses: false
:exclude_tests: [:serializers]
...
```

drwl:
No this was my fault with bad instructions in the top PR message. Psych/YAML parser does not have a concept of Ruby symbols so you have to remove the leading colon:

```yaml
...
:exclude_fixtures: false
:exclude_helpers: true
:exclude_scaffolds: true
:exclude_serializers: false
:exclude_sti_subclasses: false
:exclude_tests: [serializers] # alternatively, ['serializers'] 
...
```

drwl:
Also note that you want to pass the singular (non-plural) form to exclude_tests, so it should be:

`:exclude_tests: [serializer]`

JordanFinch72:
Works really nicely! I would have maybe expected it to remove existing annotations after the options changed, but perhaps that's an implementation detail you've already considered and decided against. I see there's an `annotaterb models -d` command - perhaps that could be extended to the other types of files that can be annotated at some point? :) 

Either way, great stuff! Thanks for your work on this.

drwl:
> Works really nicely! I would have maybe expected it to remove existing annotations after the options changed, but perhaps that's an implementation detail you've already considered and decided against. I see there's an `annotaterb models -d` command - perhaps that could be extended to the other types of files that can be annotated at some point? :)

I think you can achieve deleting old annotations by:
- Changing the exclusion/inclusion in the config
- Running annotaterb models --delete
- Adding back the exclusion config and then rerunning migrations

> Either way, great stuff! Thanks for your work on this.

Thanks and likewise. I'm going to merge this in and push a new minor which will have this change.

PR #109, opened 2024-04-30
Fix misspelling in migration guide

RobinDaugherty:
[The past tense of "lead" is "led"](https://www.merriam-webster.com/grammar/2281)


drwl:
Apologies for the delay in response — thanks for catching this and putting up a fix!

PR #118, opened 2024-05-28
Add support for `NOT VALID` constraints

gmcabrita:
Thank you for forking annotate_models, cleaning up the code, and providing further maintenance for it 🙏 

We were about to swap our extremely old fork with check constraint support to annotaterb and noticed that `NOT VALID` constraints are not fully annotated.

For example, one of our model annotations would end up missing the `NOT VALID` portion after migrating:
```diff
diff --git a/app/models/model_name.rb b/app/models/model_name.rb
index f39ff14a38..c7220921cb 100644
--- a/app/models/model_name.rb
+++ b/app/models/model_name.rb
@@ -34,10 +34,10 @@
 #
 #  fk_rails_...  (foreign_table_id => foreign_table_name.id)
 #
-# CHECK Constraints
+# Check Constraints
 #
-#  positive_check  ((column_a > 0)) NOT VALID
-#  length_check    ((length(column_b) <= 100000)) NOT VALID
+#  positive_check  (column_a > 0)
+#  length_check    (length(column_b) <= 100000)
 #

 class ModelName < ApplicationRecord
```

So I went ahead and added support for it.

I'm not sure what the oldest Rails version you want to support is, but it seems that `validated?` is [available from Rails 5.2.2 onwards](https://github.com/rails/rails/commit/8203482a9ef9bd60d5014745fd7af868d9954b1d#diff-fac852e063eee718c2af44dec6f80b1b46f52f8aae7936e5eea6b2eab08cf2f7R90-R94).

### DIFF
--- a/lib/annotate_rb/model_annotator/check_constraint_annotation/annotation_builder.rb
+++ b/lib/annotate_rb/model_annotator/check_constraint_annotation/annotation_builder.rb
@@ -24,7 +24,13 @@ module AnnotateRb
 
           max_size = check_constraints.map { |check_constraint| check_constraint.name.size }.max + 1
           check_constraints.sort_by(&:name).each do |check_constraint|
-            expression = check_constraint.expression ? "(#{check_constraint.expression.squish})" : nil
+            expression = if check_constraint.expression
+              not_validated = if !check_constraint.validated?
+                "NOT VALID"
+              end
+
+              "(#{check_constraint.expression.squish}) #{not_validated}".squish
+            end
 
             constraint_info += if @options[:format_markdown]
               cc_info_in_markdown(check_constraint.name, expression)
--- a/spec/lib/annotate_rb/model_annotator/annotation_builder_spec.rb
+++ b/spec/lib/annotate_rb/model_annotator/annotation_builder_spec.rb
@@ -2016,6 +2016,35 @@ RSpec.describe AnnotateRb::ModelAnnotator::AnnotationBuilder do
         is_expected.to eq expected_result
       end
 
+      context "when constraint is not valid" do
+        let :custom_connection do
+          check_constraints = [
+            mock_check_constraint("must_be_us_adult", "age >= 18", false)
+          ]
+
+          mock_connection([], [], check_constraints)
+        end
+        let :expected_result do
+          <<~EOS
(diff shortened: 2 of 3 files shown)


drwl:
Hi @gmcabrita thanks for the kind words and submitting a PR. I'm trying to better understand what the change is doing, it looks like we want to add "NOT VALID" string to check constraints if they're not validated. Could you elaborate on what it means to have a not-validated check constraint?

drwl:
I'm able to see here [1] it's a postgres option only, but in my mind I'm trying to understand the use case 

[1] https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_check_constraint

gmcabrita:
Yes, it's a Postgres only option. The ActiveRecord class assumes it is valid by default, so `validated?` should always return `true` for other databases like MySQL.

The `NOT VALID` constraints in Postgres are mainly useful for two use cases:
1. Adding a foreign key or check constraint with zero downtime
	a. Normally, adding one of these constraints requires a full table scan to verify that all existing rows satisfy the constraint. This operation [can cause problems in production](https://github.com/ankane/strong_migrations#adding-a-check-constraint) because it blocks all writes to the table. For this reason, constraints are first added as `NOT VALID` (only inserts/updates are validated by the constraint) and later validated in a separate migration.
2. Choosing to validate the constraint only against new inserts/updates
	a. This is very rare in my experience.

Having this be reflected in the annotations is valuable because it can either signal that something is wrong (e.g. we forgot to add the migration to validate the constraint), or that the constraint exists but we should not expect it to hold valid for all rows in the table.

This is also what you'd see in the schema dump:
```sql
ALTER TABLE public.table_name
	ADD CONSTRAINT length_check CHECK ((length(column_b) <= 100000)) NOT VALID;
```

drwl (approved):

drwl:
@gmcabrita thanks for the explanation. I learned a bit more about Rails + Postgres. I'll get this merged in and cut a new version.

drwl:
@gmcabrita should be live with 4.9.0 that just got pushed to Rubygems

gmcabrita:
@drwl Thank you for the quick turnaround 🫶

PR #124, opened 2024-06-22
Annotate routes on DB tasks to match the old annotate gem behavior

AliOsm:
Thanks for creating this new annotate gem. I propose to also run `routes` command after `models` command within DB tasks to match the old annotate gem behavior. I think it is easier to run `rails db:migrate` after each change in a model or `routes.rb`, than remembering to run either `rails db:migrate` or `annotaterb routes`.

### DIFF
--- a/lib/annotate_rb/tasks/annotate_models_migrate.rake
+++ b/lib/annotate_rb/tasks/annotate_models_migrate.rake
@@ -30,6 +30,7 @@ migration_tasks.each do |task|
 
     Rake::Task[task_name].enhance do
       ::AnnotateRb::Runner.run(["models"])
+      ::AnnotateRb::Runner.run(["routes"])
     end
   end
 end


drwl:
> Thanks for creating this new annotate gem. I propose to also run `routes` command after `models` command within DB tasks to match the old annotate gem behavior. I think it is easier to run `rails db:migrate` after each change in a model or `routes.rb`, than remembering to run either `rails db:migrate` or `annotaterb routes`.



> Thanks for creating this new annotate gem. I propose to also run `routes` command after `models` command within DB tasks to match the old annotate gem behavior. I think it is easier to run `rails db:migrate` after each change in a model or `routes.rb`, than remembering to run either `rails db:migrate` or `annotaterb routes`.

drwl:
Hi @AliOsm thanks for using the gem and adding a PR. Could you tell me more about your use case so that I can better understand the pain point?

From what I'm reading, it seems like there's enough changes to routes that they go stale, and that by coupling the db task hook it solves your pain point.

AliOsm:
Hello @drwl, this is exaclty the pain point, when I change the routes file, I always forget to update the annotations. Maybe coupling the update with each `db:migrate` is overkill, but maybe adding an option like `annotate_routes_on_db_tasks`, and if it is `true`, then routes will be annotated on `db:migrate`.

drwl:
> Hello @drwl, this is exaclty the pain point, when I change the routes file, I always forget to update the annotations. Maybe coupling the update with each `db:migrate` is overkill, but maybe adding an option like `annotate_routes_on_db_tasks`, and if it is `true`, then routes will be annotated on `db:migrate`.

Got it, I think I understand. I'm not sure if there's a lot of demand from other folks for coupling this, so for the time being I don't think it makes sense to go with this change at a global level.

Just as a reminder, if you run the generator to add the db task hook (`bin/rails g annotate_rb:hook`), you can add the desired change to the copy that gets installed in your `lib/tasks` directory.

AliOsm:
I agree with you, and copying the hook is a good option, but it will require the user to manually update it in future, right?

If you are ok with the new option solution mentioned above, I can try to implement it, let me know please.

drwl:
> I agree with you, and copying the hook is a good option, but it will require the user to manually update it in future, right?
> 
> If you are ok with the new option solution mentioned above, I can try to implement it, let me know please.

Just to clarify, so the hook gets installed using the rails generator command. Assuming there are changes in the future (to the hook), then yes it would require the user to update the hook and re-add the line.

I'm not convinced there's enough demand to warrant such a change at a global scale. From periodically looking at issues on the old gem, it doesn't seem like a common use case. I'm open to revisiting this in the future, but I think for the time being it might be best to make the changes to your local hook file with the 1 liner change.

AliOsm:
That's fine, thanks for the discussion, I will close this PR.

drwl:
No worries. I'm happy to revisit this in the future if the change makes sense.

PR #131, opened 2024-07-03
Correct uses of `respond_to` in AnnotationDecider

andreccosta:
In our project we have more than just Active Record models within our `models` folder. After switching to this gem we noticed a couple errors being raised when attempting to decide if some of those should be annotated or not.

For example, a class that inherits from `Dry::Struct` responds to `abstract_class` but not to `abstract_class?`.

```ruby
> class Demo < Dry::Struct; end
> Demo.abstract_class
=> Dry::Struct
> Demo.abstract_class?
(irb):3:in `<main>': undefined method `abstract_class?' for class Demo (NoMethodError)
```

### DIFF
--- a/lib/annotate_rb/model_annotator/annotation_decider.rb
+++ b/lib/annotate_rb/model_annotator/annotation_decider.rb
@@ -23,8 +23,8 @@ module AnnotateRb
           return false if !klass_is_a_class
 
           klass_inherits_active_record_base = klass < ActiveRecord::Base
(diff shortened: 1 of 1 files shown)


drwl:
@andreccosta thanks for trying out this gem and submitting a PR, it's very much appreciated.

Changes look good to me!

drwl:
@andreccosta just a heads up, I published v4.10.1 to Rubygems and it includes this fix.

andreccosta:
@drwl thank you!

PR #135, opened 2024-07-08
Fix related files in multi-db scenario

drwl:


drwl:
Taken from https://github.com/ctran/annotate_models/pull/892

> Given a Rails application with multiple databases, tables in two databases with the same table name, and corresponding models with different names, some files get annotated with the wrong table's schema.
> 
> The issue is that some of the patterns include `%TABLE_NAME%`, which will be identical between the two tables, even though they have different model names.
> 
> This fixes the issue by eliminating the use of the table name when the model is not connected to the primary database.

drwl:
The issue lies in building the list of related files for a model file. Using the example Rails project https://github.com/cygnuseducation/annotate-test-app, running `bundle exec annotaterb models` consecutively led to the same output. This happened because the factory file `test/factories/users.rb` would get annotated using `Secondary::User`'s annotations and then `User`'s annotations.

```
[6] pry(#<AnnotateRb::ModelAnnotator::ProjectAnnotator>)> c
file_name: app/models/secondary/user.rb called
file_name: test/factories/users.rb called
file_name: test/factories/secondary/users.rb called
file_name: app/models/user.rb called
file_name: test/factories/users.rb called
Annotated (2): test/factories/users.rb, test/factories/users.rb
```

drwl:
Note for any future viewers, the change may require Rails 6+ support as the behavior for `#connection_specification_name` changed between Rails 5 and Rails 6.1.

https://apidock.com/rails/v5.0.0.1/ActiveRecord/ConnectionHandling/connection_specification_name

https://apidock.com/rails/v6.1.3.1/ActiveRecord/ConnectionHandling/connection_specification_name

drwl:
I currently don't have time for this, so if any future viewer wants to take this on feel free. Closing this PR for now.

PR #136, opened 2024-07-12
Change structure of model annotation builder

drwl:
This is a big PR in terms of changes but does the following:
- builds on top of #134, by changing how `ModelAnnotation::Annotation::AnnotationBuilder` interacts with component classes
- makes it easier to understand how the model annotations are built
- breaks down big test cases into smaller unit tests and placed in their appropriate places


drwl:
While making these changes, I realized that index information gets used in 2 places: in the column annotations and index annotation. Also, some logic is hidden away in `ModelWrapper` and `ColumnWrapper`.

PR #137, opened 2024-07-13
Allow for separating associations into their own section in the annotation

carldr:
Adds a `--separate-associations` option which puts associations into their own section.  This allows for easier eyeballing of associations in the table, even more so than `--classified-sort`.  Tests and updated README included.

Produces output like :

```
# == Schema Information
#
# Table name: team_users
#
#  id         :bigint           not null, primary key
#  roles      :jsonb            not null
#
# Associations
#
#  team_id    :bigint           not null
#  user_id    :bigint           not null
#
# Indexes
#
#  index_team_users_on_team_id  (team_id)
#  index_team_users_on_user_id  (user_id)
#
# Foreign Keys
#
#  fk_rails_6a8dc6a6fc  (team_id => teams.id)
#  fk_rails_8b0a3daf0d  (user_id => users.id)
#
```

The pull request also includes a change to add a period between the table name and composite keys.  Previously it would produce :

```
#  fk_rails_da54bff5b7  ([team_id, cheese_id] => cheeses[team_id, id])
```

And now :

```
#  fk_rails_da54bff5b7  ([team_id, cheese_id] => cheeses.[team_id, id])
```

which better matches the formatting of non-composite keys.

And lastly, a change to gemspec which allows the Gem to be used locally with `bundle config set local.annotaterb <local-path>` inside a Rails app, which avoids the need to commit and push inside annotaterb, then bundle update in the Rails app, so see how changes affect real-world output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ Additional options that work for annotating models and routes
         --ignore-model-subdirects    Ignore subdirectories of the models directory
         --sort                       Sort columns alphabetically, rather than in creation order
         --classified-sort            Sort columns alphabetically, but first goes id, then the rest columns, then the timestamp columns and then the association columns
+        --separate-associations      Puts associations in their own section
     -R, --require path               Additional file to require before loading models, may be used multiple times
     -e [tests,fixtures,factories,serializers],
         --exclude                    Do not annotate fixtures, test files, factories, and/or serializers
--- a/annotaterb.gemspec
+++ b/annotaterb.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |spec|
   spec.name = "annotaterb"
-  spec.version = File.read("VERSION").strip
+  spec.version = File.read(__dir__ + "/VERSION").strip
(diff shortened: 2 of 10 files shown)


carldr:
Added a commit to fix some style issues, I'd missed the standardrb step.

drwl:
Hi @carldr, thanks so much for submitting a PR. I think the changes make sense to me, I'm going to spend a bit more time thinking through the separate associations bit.

Would you mind splitting up the changes into separate PRs so that it's easier for any future person to parse the changes? Specifically, the separate associations change, changes to gemspec, and the change to composite foreign keys?

carldr:
> Would you mind splitting up the changes into separate PRs so that it's easier for any future person to parse the changes? Specifically, the separate associations change, changes to gemspec, and the change to composite foreign keys?

Yeah, will do, give me a day or two to do that.

drwl:
I'm doing a clean up of the project PRs and issues. Closing this for now as it's a bit stale. Feel free to re-open if/when you get around to this. Happy to help on the review side.

PR #139, opened 2024-07-22
bootstrap: only load Rakefile if needed [wip]

dmke:


dmke:
Sorry, wrong base repository. This is still heavily work-in-progress.

You'll find the continuation here: https://github.com/dmke/annotaterb/tree/fix2/rake

Issue #130, opened 2024-06-28
Rakefile seems to be loaded twice


dmke:
I have some Rake files, which define some constants. When running `bin/rails db:migrate`, the output is flooded with warnings of already defined constants.

This doesn't happen when calling `bundle exec annotaterb models`.

## Commands

```
$ bin/rails db:migrate
RAILS_ROOT/lib/tasks/codegen.rake:48: warning: already initialized constant Codegen::RAILS_ROOT
RAILS_ROOT/lib/tasks/codegen.rake:48: warning: previous definition of RAILS_ROOT was here
...
Annotating models
Model files unchanged.
```

## Reproduction

- Start with an empty Rails project (`rails new --minimal --api` or similar)
- Create a model (`rails g model User name email:unique`) (not sure if this is necessary)
- Create a `lib/tasks/codegen.rake` (the actual name is not relevant, it must end in `.rake` though):
  ```ruby
  module Codegen
    RAILS_ROOT = Pathname.new(__dir__).join("../..").expand_path

    # imagine some code generation methods here
  end

  namespace :codegen do
    task :ts_types do
      # imagine calling Codegen.ts_types or similar here
    end
  end
  ```
- Create/migrate database (`bin/rails db:{create,migrate}`). Observe no warning.
- Add `annotaterb`, configure `skip_on_db_migrate: false`.
- Migrate again (`bin/rails db:migrate`). Observe warnings.

## Breakdown

Adding a `puts caller` before the `Codegen::RAILS_ROOT` constant reveals the following:

```
--- first encounter ---
GEM_HOME/gems/railties-7.1.3.4/lib/rails/engine.rb:684:in `load'
GEM_HOME/gems/railties-7.1.3.4/lib/rails/engine.rb:684:in `block in run_tasks_blocks'
GEM_HOME/gems/railties-7.1.3.4/lib/rails/engine.rb:684:in `each'
GEM_HOME/gems/railties-7.1.3.4/lib/rails/engine.rb:684:in `run_tasks_blocks'
GEM_HOME/gems/railties-7.1.3.4/lib/rails/application.rb:583:in `run_tasks_blocks'
GEM_HOME/gems/railties-7.1.3.4/lib/rails/engine.rb:471:in `load_tasks'
RAILS_ROOT/Rakefile:6:in `<top (required)>'
GEM_HOME/gems/rake-13.2.1/lib/rake/rake_module.rb:29:in `load'
[...]

PR #140, opened 2024-07-22
Fix double-loading of Rakefile

dmke:
See #130 for problem discussion.

This also drops a lot of legacy and/or unnecessary code from `AnnotateRb::RakeBootstrapper`.


drwl (approved):
Thanks so much for adding these changes!

PR #163, opened 2024-11-14
Add support for virtual columns

robbevp:
Hi there 👋 

Thanks for this active fork! I had a personal fork with some changes for the longest time.

One of these changes was support for [virtual/generated columns](https://guides.rubyonrails.org/active_record_postgresql.html#generated-columns). 
At the moment, the following table would create a slightly confusing annotation
```ruby
create_table :users do |t|
  t.string :name
  t.virtual :name_upcased, type: :string, as: "upper(name)", stored: true
end
```

```ruby
# Table name: users
#
#  name                :string
#  name_upcased :string
```

With these changes, this would generate the following:
```ruby
# Table name: users
#
#  name                :string
#  name_upcased :virtual(string)
```

I wasn't a 100% sure where you'd want these changes exactly. I'm happy to make further changes, or add some more tests.

One additional change could be to also place the function in the annotation, but this would maybe better in a further PR. I've tried this in the past and noticed that it can become unwieldy once the function becomes more complicated
```ruby
# Table name: users
#
#  name.                :string
#  name_upcased :virtual(string)    upcase(name)
```

### DIFF
--- a/lib/annotate_rb/model_annotator/column_annotation/attributes_builder.rb
+++ b/lib/annotate_rb/model_annotator/column_annotation/attributes_builder.rb
@@ -78,6 +78,12 @@ module AnnotateRb
             end
           end
 
+          # Check if the column is a virtual column and print the function
+          if @options[:show_virtual_columns] && @column.virtual?
+            # Any whitespace in the function gets reduced to a single space
+            attrs << @column.default_function.gsub(/\s+/, " ").strip
+          end
+
           attrs
         end
 
--- a/lib/annotate_rb/model_annotator/column_annotation/column_wrapper.rb
+++ b/lib/annotate_rb/model_annotator/column_annotation/column_wrapper.rb
@@ -85,6 +85,10 @@ module AnnotateRb
           @column.name
         end
 
+        def virtual?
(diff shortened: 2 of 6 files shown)


drwl:
Hi @robbevp thanks for putting up a PR. I don't have any objections here, but let me play around with this. 

At a high level, my only concern is that adding this without an option flag _could_ affect others.

drwl:
I had a chance to play around with this locally. The direction seems fine. I'm curious what it looks like with the functions as well.

I'd be okay taking the current changes, but would prefer there be a corresponding option flag, perhaps something like `show_virtual_columns`? I'm open to a different/better naming.

robbevp:
Hi @drwl 
I added the option - I couldn't think of anything better, so went with your suggestion.
Since we now have an option, I also added a separate commit that will print the function, so you can try this as well. I'm ok with dropping that commit if you don't want those changes (yet).

robbevp:
Just saw that I forgot to lint my files yesterday, this is fixed now

robbevp:
@drwl Any chance you could take another look at this?

drwl (approved):
Looks good to me 👍, thanks for adding this!

PR #173, opened 2024-12-31
feat: add `timestamp_columns` config option

pbernays:
I'm working with a codebase that doesn't use Rails' default timestamp columns, so being able to customise those would be helpful. This adds `timestamp_columns` as a configurable option for classified sorting, and sorts them in config order rather than alphabetical order.


pbernays:
I'd prefer this one was independent. It's only in the other PR because I messed up organising my branches. The other PR was and is intended to be food-for-thought, and probably something that needs more TLC before it gets merged while I try to align with your goals.

drwl (approved):
Thank you for this and will cut a release for this soon

PR #174, opened 2024-12-31
chore: ensure Options are initialised with defaults

pbernays:
I can't see a reason not to load defaults when initialising an Options object (and the comment on line 158/96 suggests it's desired behaviour), so this calls `load_defaults` during init and makes it a private method.

Correct me if I'm wrong, but I also can't see a reason to keep the unused `format_bare` flag option given that arbitrary keys are allowed.

Removes the `*_OPTIONS_KEYS` constants for being redundant.

Tidies up a couple of lines to be clearer.

### DIFF
--- a/lib/annotate_rb/options.rb
+++ b/lib/annotate_rb/options.rb
@@ -10,7 +10,7 @@ module AnnotateRb
 
     class << self
       def from(options = {}, state = {})
-        new(options, state).load_defaults
+        new(options, state)
       end
     end
 
@@ -36,7 +36,6 @@ module AnnotateRb
       exclude_sti_subclasses: false, # ModelAnnotator
       exclude_tests: false, # ModelAnnotator
       force: false, # ModelAnnotator, but should be used by both
-      format_bare: true, # Unused
       format_markdown: false, # ModelAnnotator, RouteAnnotator
       format_rdoc: false, # ModelAnnotator
       format_yard: false, # ModelAnnotator
@@ -92,71 +91,10 @@ module AnnotateRb
 
     DEFAULT_OPTIONS = {}.merge(POSITION_OPTIONS, FLAG_OPTIONS, OTHER_OPTIONS, PATH_OPTIONS).freeze
 
-    FLAG_OPTION_KEYS = [
-      :classified_sort,
-      :exclude_controllers,
-      :exclude_factories,
-      :exclude_fixtures,
-      :exclude_helpers,
-      :exclude_scaffolds,
-      :exclude_serializers,
-      :exclude_sti_subclasses,
-      :exclude_tests,
-      :force,
-      :format_markdown,
-      :format_rdoc,
-      :format_yard,
-      :frozen,
-      :ignore_model_sub_dir,
-      :ignore_unknown_models,
-      :include_version,
-      :show_check_constraints,
-      :show_complete_foreign_keys,
-      :show_foreign_keys,
-      :show_indexes,
-      :simple_indexes,
-      :sort,
-      :timestamp,
-      :trace,
-      :with_comment,
-      :with_column_comments,
-      :with_table_comments
-    ].freeze
-
-    OTHER_OPTION_KEYS = [
-      :active_admin,
-      :command,
-      :debug,
-      :hide_default_column_types,
-      :hide_limit_column_types,
-      :ignore_columns,
-      :ignore_routes,
-      :ignore_unknown_models,
-      :models,
-      :routes,
-      :skip_on_db_migrate,
-      :target_action,
-      :wrapper,
-      :wrapper_close,
-      :wrapper_open,
-      :classes_default_to_s
-    ].freeze
-
-    PATH_OPTION_KEYS = [
-      :additional_file_patterns,
-      :model_dir,
-      :require,
-      :root_dir
-    ].freeze
-
-    ALL_OPTION_KEYS = [
-      POSITION_OPTIONS.keys, FLAG_OPTION_KEYS, OTHER_OPTION_KEYS, PATH_OPTION_KEYS
-    ].flatten.freeze
-
     POSITION_DEFAULT = "before"
 
     # Want this to be read only after initializing
-    def_delegator :@options, :[]
+    def_delegators :@options, :[], :to_h
 
     def initialize(options = {}, state = {})
       @options = options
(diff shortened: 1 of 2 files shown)


drwl:
Hi @pbernays thanks for submitting a PR and apologies on getting around to this. 

I'll try to respond to this inline -- and to be clear, I'm happy to have a conversation around this so that we can improvement the gem.

> I can't see a reason not to load defaults when initialising an Options object (and the comment on line 158/96 suggests it's desired behaviour), so this calls load_defaults during init and makes it a private method.

To provide some history, this gem is a hard fork from `annotate_models`. One of the biggest issues in the old gem was that configuration options and state were global. They could be set in multiple places, and were also read from different places.

One of my initial goals was to make it easier to have isolated functions that were easy to unit test. To get that, most functions now take an `AnnotateRB::Options` as a parameter and read from that, instead of a global variable. Something I found while doing this, was that some of the default options would have unexpected effects on the code because I didn't explicitly set them. See https://github.com/drwl/annotaterb/pull/34

> Correct me if I'm wrong, but I also can't see a reason to keep the unused format_bare flag option given that arbitrary keys are allowed.

We can get rid of this flag -- you're right, it's not used. I didn't touch it because I wanted the migration from the old gem to this one to be easy.

> Removes the *_OPTIONS_KEYS constants for being redundant.

When I was first doing an audit of the various options, I made these hashes separate so it was clearer what was being used and where. If it doesn't make sense to keep these anymore then it seems fine to remove.

pbernays:
> Hi @pbernays thanks for submitting a PR and apologies on getting around to this.

First off, and this is overdue, I owe you thanks for your gift to open source. I wasn't looking forward to trying to alter the code of the original gem to add the changes in the other PR I put up, but a colleague showed me your 2yo reddit post, and your refactor made it a breeze. So thank you, I appreciate the time and effort you've put in to improve it.

Also, I don't believe you owe an apology either, especially when I've been sloppy and haven't corrected the rubocop nitpicks yet.

> I'll try to respond to this inline -- and to be clear, I'm happy to have a conversation around this so that we can improvement the gem.

I'd be happy to be involved in that if you'd like input from collaborators. The tool's always been a set-and-forget kind of deal for me, so I don't have much of a vision for it. But it is very useful and I'd like to contribute to keep it that way, aligned with whatever your plans for it are. Which begs the question, what are your plans?

> > I can't see a reason not to load defaults when initialising an Options object (and the comment on line 158/96 suggests it's desired behaviour), so this calls load_defaults during init and makes it a private method.
> 
> To provide some history, this gem is a hard fork from `annotate_models`. One of the biggest issues in the old gem was that configuration options and state were global. They could be set in multiple places, and were also read from different places.
> 
> One of my initial goals was to make it easier to have isolated functions that were easy to unit test. To get that, most functions now take an `AnnotateRB::Options` as a parameter and read from that, instead of a global variable. Something I found while doing this, was that some of the default options would have unexpected effects on the code because I didn't explicitly set them. See #34

Thanks for the clarity. It's been a couple of months since I had my head in this code, and I remember having some thoughts about the specs to do with the options, but what they were... ¯\\_(ツ)_/¯. It seems worth removing this change, figuring out what those ideas are to discuss with you, and then follow up if you're in agreement.

> > Correct me if I'm wrong, but I also can't see a reason to keep the unused format_bare flag option given that arbitrary keys are allowed.
> 
> We can get rid of this flag -- you're right, it's not used. I didn't touch it because I wanted the migration from the old gem to this one to be easy.

Makes perfect sense.

> > Removes the *_OPTIONS_KEYS constants for being redundant.
> 
> When I was first doing an audit of the various options, I made these hashes separate so it was clearer what was being used and where. If it doesn't make sense to keep these anymore then it seems fine to remove.

👍

lib/annotate_rb/options.rb
```diff
@@ -92,102 +94,61 @@ def from(options = {}, state = {})
...
-
-    ALL_OPTION_KEYS = [
-      POSITION_OPTIONS.keys, FLAG_OPTION_KEYS, OTHER_OPTION_KEYS, PATH_OPTION_KEYS
-    ].flatten.freeze
-
     POSITION_DEFAULT = "before"
 
     # Want this to be read only after initializing
-    def_delegator :@options, :[]
+    def_delegators :@options, :[], :to_h
```
drwl, line 100: I'd definitely like your thoughts here -- 

So when I refactored the old gem, I intentionally made `Options` read only. In the old gem, it was not clear where options were being set as it was a read-writeable global variable, which made it really hard to debug.

However, we are no longer in the old gem and so it might be less of a concern. Thoughts on keeping it the same versus making it writable by delegating `to_h`?

drwl (commented):
Everything looks good to me -- just 1 question/discussion point

lib/annotate_rb/options.rb
pbernays: That's a good question. Off the top of my head, I don't believe I wanted this to be writable but a convenience to get a copy. I can't see where I've even used it in the PR, so it might have been something I failed to cleanup after debugging in the console. I think your initial intention is good, I can't think of a reason `Options` should need to change after they've been initialized.

pbernays:
> I have some downtime and am tidying up the repo. It seems like we had some discussions but stalled out -- is this still something you want to pursue?

Sorry, life went a bit sideways this year so I lost focus here.

> Looking over the notes, I'm 50/50 if this change should go through for testing reasons. I _think_ unit testing would be easier if the options could be initialized without any side effects.

I see this Options object as an immutable object we can always query and expect to receive a meaningful value, even (or especially) when there's no external configuration. As a user of the class I'd like it to be perfectly reliable, ie. there's no pathway that allows me to put Options in an invalid state. In theory, if we test the way we use the class, making it easy to use should make it easy to test.

If that aligns with your view I'd be happy to pick it up again, otherwise no issue.

drwl:
I have been responding to PRs and issues in a stack like fashion, and am starting at the beginning of the queue for now.

After reviewing this again, I think it doesn't make sense to make this change at this time. If there's a stronger push for this then I'm happy to reevaluate the decision. 

> I see this Options object as an immutable object we can always query and expect to receive a meaningful value, even (or especially) when there's no external configuration. As a user of the class I'd like it to be perfectly reliable, ie. there's no pathway that allows me to put Options in an invalid state. In theory, if we test the way we use the class, making it easy to use should make it easy to test.

I agree that Options should be immutable, but currently the intention is that it gets set externally, in other words it's just a dummy key value store. Ideally it should be set once and then not modified (aside from the `state` that got added).

Given the expanding list of options that users of the gem add, I do think there'll be a need in the future for validations -- this would prevent mutators of Options to put it in an invalid state. 

Hopefully my reasoning makes sense. As part of repo clean up, I'm going to close this PR for now.

PR #177, opened 2025-01-16
Update README.md

kbruccoleri:
When specifically requiring the gem, the path in `lib` is `annotate_rb.rb`, with the underscore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Or install it into your Rails project through the Gemfile:
 group :development do
   ...
 
-  gem "annotaterb"
+  gem "annotaterb", require: "annotate_rb"
   
   ...
 ```


kbruccoleri:
We noticed in our code that `AnnotateRb` and all other constants were not available unless we loaded the `lib/annotate_rb` file.

I think this is because the gem name `annotaterb` does not match the path of the `main` file `lib/annotate_rb.rb`.

yskkin:
Instead of adding require instruction, I thinkg putting `lib/annotaterb.rb`, whose content is `require annotate_rb` is better approach. 
Users do not have to care about this, and is backward compatible for users who already `require annotate_rb`.

drwl:
> Instead of adding require instruction, I think putting `lib/annotaterb.rb`, whose content is `require "annotate_rb"` is better approach. Users do not have to care about this, and is backward compatible for users who already `require "annotate_rb"`.

By this do you mean just add `lib/annotaterb.rb` file to the gem?

yskkin:
> By this do you mean just add `lib/annotaterb.rb` file to the gem?

Yes, with proper content.

drwl:
I merged in #185 to fix this issue. Thank you for the help in reporting and suggesting improvements. I'm going to close this issue, but feel free to open it if it is not addressed.

PR #180, opened 2025-01-22
Lock `concurrent-ruby` gem to fix CI

tagliala:
To address compatibility issues with Rails 7.0, this commit locks `concurrent-ruby` gem to version 1.3.4.

This change is necessary to prevent an uninitialized constant error related to `ActiveSupport::LoggerThreadSafeLevel::Logger`, as documented in issue rails/rails#54260.

By specifying this version, we ensure that our CI integration specs pass consistently, providing a stable development and testing environment.

This measure will remain in place until the integration tests upgrade to Rails 7.1, which involves significant changes that require careful consideration and testing.


drwl:
Thank you very much for submitting this!

drwl (approved):

PR #181, opened 2025-01-23
Fix translation foreign key exclusion bug

galori:
I ran into a bug with this logic when using annotaterb + globalize:

For compound names for models such as "ABCustomer", it resulted in "abcustomer_id" being added to the exclusion list, where the actual foreign key name was "ab_customer_id". 

Therefore it was not excluded. And therefore all the annotations for ABCustomer ended up including "ab_customer_id" even though that wasn't an actual column.

This PR includes changes to the unit tests to not break them. However it didn't seem like this bug something I could actually reproduce with the unit tests. I couldn't get the integration test suite working and couldn't spend more time on it right now.

### DIFF
--- a/lib/annotate_rb/model_annotator/model_wrapper.rb
+++ b/lib/annotate_rb/model_annotator/model_wrapper.rb
@@ -167,7 +167,7 @@ module AnnotateRb
           :created_at,
           :updated_at,
           :locale,
-          @klass.name.foreign_key.to_sym,
+          @klass.name.foreign_key.to_sym
         ]
       end
     end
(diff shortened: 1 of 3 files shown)


drwl (changes requested):
Thanks for submitting this!

I'm not too familiar with using the globalize gem, but if this works for your app then I'm okay to merge this in. Can you just fix the linter issue then re-request review?

galori:
> Can you just fix the linter issue then re-request review?

Linter errors fixed. 🙏

drwl (approved):

PR #183, opened 2025-01-31
chore: add rake task to automatically deploy to rubygems

OdenTakashi:
Thank you always for your maintenance !

## Description
Since the tag mentioned, "Manually doing releases until an automated solution is put in place,"([Release Bump version to v4\.13\.0 \(\#159\) · drwl/annotaterb](https://github.com/drwl/annotaterb/releases/tag/v4.13.0)) 
I have added a Rake task to automatically deploy to RubyGems and generate tags. 

You can deploy to RubyGems and generate a tag by running the following command from the CLI:

```sh
$ rake release
```

I believe the following steps are involved in releasing a Gem:

- Bumping up the version
- Writing the CHANGELOG
- Creating a tag
- Pushing the tag to the remote
- Pushing to RubyGems

With the current changes, the following steps can be automated:

- Creating a tag
- Pushing the tag to the remote
- Pushing to RubyGems

However, since the description for the generated tag is not dynamic. ( [rubygems/bundler/lib/bundler/gem\_helper\.rb ](https://github.com/rubygems/rubygems/blob/c52a74962132c54681334300c7069bb852c4419f/bundler/lib/bundler/gem_helper.rb) )
So I believe it might be worth considering a Release Note if additional explanation is needed. 

If release note is needed, please let me know, and I will add workflow that creates a release note triggered by a tag.
However, we may need to discuss the wording and other details.


OdenTakashi:
@drwl
Please review this whenever it’s convenient for you.  :pray:

drwl:
@OdenTakashi thanks for adding this. I will need to research how other gem maintainers do releases and I think this could be helpful in automating some of the process.

Currently, I make a branch with the new version history so I don't commit and push to main directly. Although, I wonder if this is okay for releases?

OdenTakashi:
@drwl 
Thank you for checking 👍 
Since it's your gem, please consider this just as a reference.

> Currently, I make a branch with the new version history so I don't commit and push to main directly. Although, I wonder if this is okay for releases?

I think it's totally fine.

With rake release, the code that gets deployed is the one at the point where the tag is created. So as long as you tag the final commit on the release branch and then merge it into main, there shouldn’t be any issues.

Here’s the general flow I think makes sense:

- Create a release branch
- Update the CHANGELOG, bump the version, etc.
- Run rake release — this is when the deployment happens
- Open a PR and merge the branch into main

One potential risk, though, is if the release branch doesn’t get merged back into main, the released code would exist only on the tag and not on main (though you could still reference it via the tag).

So alternatively, you could follow this flow:

- Create a release branch
- Update the CHANGELOG, bump the version, etc.
- Open a PR and merge it into main
- Run rake release on the main branch

Also, I added a note in the PR description about which parts of the release process could be automated — feel free to take a look if you're interested.

drwl (approved):
Thank you for investigating and submitting a well detailed PR!

PR #187, opened 2025-02-06
Rename to annotate_rb

sethherr:
I don't understand why you have named this inconsistently - it's surprising and confusing. 

Some things use `annotaterb` - like the gem name and the `.annotaterb.yml` file - but everything else uses `annotate_rb`

This PR switches everything to be `annotate_rb` - making it simpler and using the Ruby naming convention.

I realize this would be a big change, if you accept it! But it seems like a worthwhile adjustment to make.

It makes it so `annotaterb.yml` files are still found, for backward compatibility. But, everyone using the gem will have to update the name.

An alternative option that would still achieve consistency would be renaming everything to `annotaterb` (i.e. `Annotaterb`). I like that less, because it doesn't make as much semantic sense - but if done that way, nobody would need to update the name of the gem. But - if you'd prefer that, let me know, I'm happy to submit a PR.

Regardless - thanks for creating this gem!


galori:
👍

drwl:
Hi @sethherr thanks for submitting this PR. It does seem like with more adoption of this gem, that people are running into issues with the name `annotaterb` not following conventions. So I apologize for any annoyances and headaches. I definitely had my share, for example, when working on the Rails generator for this gem.

I think you caught on to it, but I wanted to have the gem name be `annotaterb` but in code `Annotaterb` looks strange. 

Would you mind sharing where you're running into issues due to naming? It would help inform me on making a decision here.

sethherr:
For instance, when I wrote the PR to [annotate_models](https://github.com/ctran/annotate_models/pull/1040) to link to this gem, I initially used the incorrect form.

Without going and looking at the code (or re-reading my comment), I can’t tell you which one the correct form is for what. If I was adding this to a new project, I wouldn’t remember which commands go which way off the top of my head.

I have too many things on my mind to be able to remember which is which. Or I’m too dumb and lazy. Or this is too much cognitive load. Take your pick!

sethherr:
It seems like maybe this would be too inconvenient of a change - which I understand!

I'm going to close this and submit a PR for the switch in the other direction and make it backward compatible.

sethherr:
>  It does seem like with more adoption of this gem, that people are running into issues with the name `annotaterb` not following conventions.

@drwl why do you want this gem to be named `annotaterb` instead of `annotate_rb`? 

Not following naming conventions generates friction for everyone using the gem. Sometimes it's worth adding friction, but I don't understand what is gained in this instance. If you explained the advantages of having this gem named in a surprising way, maybe it would help me get over my frustration.

drwl:
@sethherr I've been putting off addressing this because this is new territory for me. I didn't expect the naming of a gem to get such a response.

I've taken long breaks between maintaining this, mostly due to work and lack of motivation, so I will need some time to re-familiarize myself with all the issues related to naming.

sethherr:
@drwl would you like to meet up or get on a call? I feel like chatting could help resolve this.

PR #192, opened 2025-02-18
Fix `changelog_uri` in gemspec

y-yagi:
This repository uses `main` for the default branch. This can reduce a redirect when accessing the URI.


drwl (approved):
Thank you for adding this!

PR #193, opened 2025-02-18
feat: Include postgres enums in output

schpet:


lib/annotate_rb/model_annotator/annotation/annotation_builder.rb
```diff
@@ -46,9 +47,34 @@ def build
 
           private
 
+          def enum_annotations
+            return [] unless @model.has_enums?
+
+            [
+              Components::Base.new.tap do |c|
+                def c.to_default
```
drwl, line 55: Could the methods be extracted to a separate class as to not use metaprogramming here?

Same as lines 61-62

lib/annotate_rb/model_annotator/column_annotation/annotation_builder.rb
```diff
@@ -28,6 +29,11 @@ def build
             @column.name
           end
 
+          if @enum_info
```
drwl, line 32: Could we also add an option flag for these changes? This would allow people currently using postgres + enums to opt-in if they want these changes, and to not introduce new unwanted changes.

drwl (changes requested):
Finally getting around to this -- first pass functionally makes sense. A few asks --

schpet:
@drwl yeah for sure, when i have time i look forward to shipping this up! thanks for the review and feedback.

PR #198, opened 2025-02-27
Support the glob pattern in `root_dir` and `model_dir`

### LINKED ISSUE #99: Feature request: packs-rails support
I'm trying out annotaterb on a Rails project which has packs-rails setup. The `packs/` directory doesn't get automatically picked up, understandably.

1. I tried to use `:root_dir: ['.', 'packs/*']` and only `app/` directory in the root gets picked up, none of the packs were recognized.
    -  I also tried variations of `packs/*` by expanding it to `packs/packone` `packs/packtwo` etc, still doesn't work.
2. I tried to use `:model_dir: ['app/models', 'packs/*/app/models'] and still the packs aren't picked up. However, expanding the glob works!

While the title is specifically on packs-rails, I wouldn't mind having a compromise via `:root_dir:` or `:model_dir:` if it supports globs 😄 

P.S. 👋🏻 hey Andrew, hope you've been well!

## Versions

- annotaterb 4.6.0
- rails 6.1
- ruby 3.1

sinsoku:
We often want to specify multiple directories in `root_dir` and `model_dir`.
For example, when using engines, packs-rails, etc.

* https://github.com/rubyatscale/packs-rails
* https://guides.rubyonrails.org/engines.html

This commit supports the glob pattern with these options, making it easy
to specify multiple directories.

fixes #99


sinsoku:
@drwl 
CI failed, so I fixed `Style/StringLiterals` and force-pushed.

drwl:
@sinsoku just checking -- have you been able to try these changes in your local? I don't have a Rails app with packwerk setup to test.

drwl (approved):
Changes lgtm -- waiting on confirmation of manual testing before merging

sinsoku:
@drwl I have tested it in a Rails app using packwerk.
It worked with this change.

jarredhawkins:
Also successfully tested this in a large Packwerked application. Thank you both for the contributions here!

PR #199, opened 2025-03-03
Place column comments at the end of the line

### LINKED ISSUE #164: Place column comments at the end of the line (feature suggestion)
## Prelude

This feature was implemented by me in [annotate_models](https://github.com/ctran/annotate_models) in those PRs: [#988](https://github.com/ctran/annotate_models/pull/988) and [#999](https://github.com/ctran/annotate_models/pull/999). Since that other GEM does not seem to be maintained anymore, I am considering a switch to `annotaterb` but would love that feature to follow.

I volunteer for implementing it, but I prefer to first post this as an issue, in case there is resistance to the idea.

It would be anyways an opt-in and would not change the behavior otherwise. It would also open the door to further formats in the future.

## Feature Suggestion

At the moment, when we configure the column comments to be documented, we get them with the name of the column, on the left of the table.

```rb
#  type                                                                                                            :enum             not null
#  book_id                                                                                                         :uuid             not null, indexed
#  parent_id(A null parent means it is a root register.)                                                           :uuid             indexed
#  starts_at(Opening date of the register (eg: for accounts, but not for categories).)                             :date
#  expires_at(Optional expiration date of the register (eg: for a credit card).)                                   :date
#  currency_iso_code                                                                                               :string(3)        not null
```

What is proposed here is to add a new configuration that allow the user to state if they want the comment with the column name (the actual behavior, and would be the default), or at the very end of the table (on the right).

The reason for such a feature is that those comments are extra information, and pushing the whole table further to the right for having them is not ideal. Also, they could potentially be long and it's then better to scroll right to read them comments, than to read their type / nullity / etc.

The proposed configuration key would be `position_of_column_comment` and could have the following values:

| Value | Description |
| -- | -- |
| `with_name` (default) | The comment appears with next to the name, in parenthesis. |
| `rightmost_column` | The comment appears in its own column, at the far right of the table. |

Example above, configured with `rightmost_column`:

```rb
#  type               :enum             not null
#  book_id            :uuid             not null, indexed   A null parent means it is a root register.
#  parent_id          :uuid             indexed
#  starts_at          :date                                 Opening date of the register (eg: for accounts, but not for categories).
#  expires_at         :date                                 Optional expiration date of the register (eg: for a credit card).
#  currency_iso_code  :string(3)        not null
```

## Contribution

As mentioned in the prelude, I offer to implement this myself, so do not hesitate to assign this to me if there is no objections.

Adeynack:
Fixes #164 

Adds the configuration `position_of_column_comment`, now defaults to `with_name` (already existing behavior), and can also be set to `rightmost_column`.

There was a bit of moving code around involved, as some information about other attribute was needed (e.g.: the `attributes` of other columns, to determine how wide this column must be).

The only formats where it got implemented though are:

- default (text)
- Markdown

The other formats I do not know well, and would ask anyone with better knowledge of them to work on it.


Adeynack:
Either the lint rules changed, or what was merged into this branch is not properly formatted. I will fix this this week.

drwl:
I just release v4.17.0 and it contains this change. Let me know if there are any issues.

PR #207, opened 2025-03-31
fix: use model name for file retrieval when not connected to the primary DB.

### LINKED ISSUE #206: Incorrect annotation of fixture files when different models share the same table name in different databases
This issue is part of the work for issue #188.

> Given a Rails application with multiple databases, tables in two databases with the same table name, and corresponding models with different names, some files get annotated with the wrong table's schema.
Ref: [\#891](https://github.com/ctran/annotate_models/issues/891)

OdenTakashi:
When the same table name exists in different databases, attempting to retrieve fixture and factory files references them by the table name. As a result, in addition to the fixtures and factories of models connected to the secondary database, related files from the primary database are also retrieved.

Therefore, this commit modifies the behavior so that for models connected to the secondary database, related files are retrieved based on the model name instead of the table name.

fixes #206


lib/annotate_rb/model_annotator/project_annotator.rb
```diff
@@ -41,7 +41,11 @@ def build_instructions_for_file(file)
         klass.reset_column_information
         annotation = Annotation::AnnotationBuilder.new(klass, @options).build
         model_name = klass.name.underscore
-        table_name = klass.table_name
+
+        # In multi-database configurations, it is possible for different models to have the same table name but live
+        #   in different databases. Here we are opting to use the table name to find related files only when the model
+        #   is using the primary connection.
+        table_name = klass.table_name if klass.connection_specification_name == ActiveRecord::Base.name
```
OdenTakashi, line 48: The behavior of the connection_specification_name method differs in Rails versions prior to 6.0.x, but since Rails 6.0.x reached EOL on June 1st, 2023, I believe there's no need to support it.
- https://apidock.com/rails/v6.0.0/ActiveRecord/ConnectionHandling/connection_specification_name
- https://apidock.com/rails/v6.1.3.1/ActiveRecord/ConnectionHandling/connection_specification_name

drwl: What happens if `klass.connection_specification_name != ActiveRecord::Base.name`, does `table_name` still need to get defined elsewhere?

OdenTakashi: As long as you're following the naming conventions, I don't think it's necessary to define table_name elsewhere.
The reason is that it is used only for file lookup, and if table_name is nil during the lookup, a search based on model_name is also performed.

https://github.com/drwl/annotaterb/blob/df91a4e53836cd9569d3984c4170f5e4b6614dd5/lib/annotate_rb/model_annotator/file_name_resolver.rb#L16

drwl: I will take another look this weekend, sorry for the delay. It seems like if we add this then the ActiveRecord version minimum will need to be bumped.

drwl (approved):
Thank you for fixing and sorry for the late review

OdenTakashi:
Please don't worry about it. I appreciate your continued maintenance and support.

PR #208, opened 2025-04-01
fix: Handle case when table_name_prefix specified as symbol

gururuby:
During of annotation of indexes I have received exception (`'String#sub': wrong argument type Symbol (expected Regexp) (TypeError)`) when some of my AR models use `table_name_prefix` option with Symbol value. In this PR I fixed it and added some specs to cover it by tests


drwl (approved):
Thanks for adding this!

PR #211, opened 2025-04-12
show included columns in indexes

pineman:


drwl:
@pineman do you know if `include` are widely used? If so, could you add a CLI and config option for this so that people currently the gem can opt into this? Right now, there's not an easy way for people to specifically choose how annotations can be formatted without these options.

pineman:
@drwl not sure exactly how widely used, but I've added an option!

drwl (approved):

PR #214, opened 2025-04-25
feat: identify unique indexes in simple_indexes option

amerritt14:
When generating simple indexes, there is no identifier for unique indexes. This PR updates the `simple_index` option to annotate with `uniquely indexed` for unique indexes.

Screenshot from a project I tested with:
image


amerritt14:
@drwl This is just an issue I found annoying, so I decided to take a crack at updating it. Love to hear if this is a desired feature.

drwl (approved):
Thanks for adding this. The change makes sense.

PR #217, opened 2025-05-08
Add debug logs for model annotation

jarredhawkins:
We have a pretty large monolith (as well as a second database host that tends to get bogged down after it runs for a while in development), so we find it useful to see some progress indicator while annotations are running.

This adds them in for each model when debug is set. This uses `puts` to match the current debug log that exists.


drwl (approved):
Thanks for adding!

PR #222, opened 2025-05-31
Fix for: Misleading pattern examples in documentation for additional_file_patterns

### LINKED ISSUE #221: Misleading pattern examples in documentation for `additional_file_patterns`
When trying to follow examples in documentation for `additional_file_patterns` to annotate [Blueprinter's](https://github.com/procore-oss/blueprinter) blueprints (serializers), which are located in `app/blueprints` directory - nothing happens. Blueprints are not picked up as additional files for annotating with `additional_file_patterns` set as `app/blueprints/%model_name%_blueprint.rb`.

### Documentation example in question
```
Additional file paths or globs to annotate, separated by commas (e.g. `/foo/bar/%model_name%/*.rb,/baz/%model_name%.rb`)
```
which does not work, because `model_name` is expected to be upper case `MODEL_NAME`.

### What does work
```
:additional_file_patterns:
- app/blueprints/%MODEL_NAME%_blueprint.rb
```

### Solution

Examples in documentation should reflect that `MODEL_NAME` in `additional_file_patterns` is expected to be upper case.

skliarov:
Fixes #221

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -56,7 +56,7 @@ Command options:
 Additional options that work for annotating models and routes
 
         --additional-file-patterns path1,path2,path3
-                                     Additional file paths or globs to annotate, separated by commas (e.g. `/foo/bar/%model_name%/*.rb,/baz/%model_name%.rb`)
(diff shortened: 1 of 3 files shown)


drwl (approved):
Thanks for reporting the issue and fixing the docs 👍

Issue #194, opened 2025-02-24
Sort by database order


pioz:
I migrated from the annotate gem, but the order in which columns are placed does not match the database order as annotate used to do. Since I use MySQL, I often run migrations with the `after:` option to insert the column at a specific point. Therefore, I would like annotaterb to also follow the column order found in the database (structure.sql). Is it possible to do this? If so, how? If not, can we add this sorting option?

pioz:
@drwl any update on this?

geremiasanti:
I have the same problem. Is there any update on the issue?

Giacomobarzon97:
fix this please

Manik5:
Same problem here, any updates?

dariot:
Fixing this would be great, now I have to launch another script that applies the correct sorting every time there's a change

hatsu38:
What is the value of classified_sort in .annotate.yml? When it is set to false, it will be the order of the columns in the DB.

It has the same behavior as classified-sort in annotate_models. `Sort columns alphabetically, but first goes id, then the rest columns, then the timestamp columns and then the association columns`
FYI: https://github.com/ctran/annotate_models?tab=readme-ov-file#options

pioz:
> Hi, I'm going through old notifications and saw this. @pioz you marked this as completed and I just wanted to confirm that it has been fixed. Is that correct?

Yes, I confirm.

PR #224, opened 2025-06-03
Set classified-sort option to false by default as it caused confusion

hatsu38:
The classified-sort option was previously set to true by default, which led to confusion for many users. It wasn't an intentional decision, so we're updating the default to false to avoid unexpected column ordering.

Ref: https://github.com/drwl/annotaterb/issues/194


drwl:
Hi @hatsu38 thank you for submitting a PR.

Can you help me understand this? In the old gem [1], was the default option for `classified_sort` false?

I _think_ it's setting it to `true` by default in these 2 places:
https://github.com/ctran/annotate_models/blob/5d01c4171990c4fe7b9b0977b05ce14a98209e53/lib/tasks/annotate_models.rake#L46
https://github.com/ctran/annotate_models/blob/5d01c4171990c4fe7b9b0977b05ce14a98209e53/lib/generators/annotate/templates/auto_annotate_models.rake#L51

[1] https://github.com/search?q=repo%3Actran%2Fannotate_models%20classified_sort&type=code

drwl:
Okay I think I understand now. I think we want to keep the CLI options to be the same to preserve the same functionality, but we can change the default here: `lib/annotate_rb/options.rb:29`

drwl (approved):
Thank you for investigating and helping me better understand the issue!

hatsu38:
I'll fix the tests that are currently failing later!

PR #225, opened 2025-06-12
Pass .annotaterb.yml through ERB

fxn:
Hi!

This PR proposes passing `.annotaterb.yml` through ERB, as Rails does with `config/database.yml` or fixtures, for instance.

Real use case is the ability to compute `model_dir` dynamically. As you know, in `annotate` the configuration can be expressed in Ruby, and therefore it can be as dynamic as you need to. But if you migrate to `annotaterb`, that is lost.

I wrote this patch on behalf of a client of mine (Gusto), whose main application has more than two hundred model dirs. They are now hard-coded in `.annotaterb.yml`, but you need to maintain that list, verify consistency in CI, etc. On the other hand, if you could compute the list in the config file itself, you could forget about it (as it was the case with `annotate`).

I have created a test suite for `load_config`, and documentation is also updated (README and migration guide). The variable `yaml_code` is named after the parameter in the signature of the called method.

The CHANGELOG seems more ad-hoc, maybe you prefer to modify this one? Please just tell me if you'd like me to do it anyway :).

### DIFF
--- a/MIGRATION_GUIDE.md
+++ b/MIGRATION_GUIDE.md
@@ -88,6 +88,16 @@ This change was done to reduce complexity in configuration and make the gem easi
 
 ----------
 
+### Dynamic configuration settings
+
+`.annotaterb.yml` is passed through ERB. Therefore, if your previous settings had dynamic values, you can still replicate that logic in the new configuration file:
+
+```yaml
+model_dir: <%= Dir["packs/*/app/models"].inspect %>
+```
+
+----------
+
 ### Automatic annotations after running database migration commands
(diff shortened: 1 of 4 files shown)


drwl:
Hi @fxn thanks for submitting this. Happy to help get this merged this in, seems like there might be an unrelated 2.7 error and then just needs linting via standard.

fxn:
Thanks @drwl! Branch updated and standard is green now.

drwl:
Just fyi, I released 4.16.0 and it should contain this change.

fxn:
@drwl that's awesome, thank you!

PR #231, opened 2025-06-18
Refactor column ignore to use `match?`

tagliala:
Switch from `match` to `match?` when filtering out ignored columns, as only the boolean result is needed.

This improves clarity and performance by avoiding unnecessary creation of `MatchData` objects.


drwl (approved):
Thanks for improving this 👍

PR #238, opened 2025-07-17
Fix classified_sort to group polymorphic association columns

garriguv:
## Summary

This PR fixes the `classified_sort` option to properly group polymorphic association columns together. Previously, polymorphic `_type` columns were sorted with regular columns while their corresponding `_id` columns were in the associations section.

## The Problem

When a model has a polymorphic association (e.g., `artifact_id` and `artifact_type`), the `classified_sort` option was not co-locating these columns. The `_type` column was treated as a regular column while the `_id` column was correctly identified as an association.

## The Solution

Modified the `classified_sort` method in `ModelWrapper` to:
1. Collect all column names in a first pass
2. Check if columns ending with `_type` have a corresponding `_id` column
3. If found, treat the `_type` column as an association column

## Example

Before this fix:
```ruby
#  id            :uuid             not null, primary key
#  amount_cents  :decimal(10, 6)
#  artifact_type :string           # <-- separated from artifact_id
#  model         :string           not null
#  provider      :string           not null
#  usage         :jsonb            not null
#  created_at    :datetime         not null
#  updated_at    :datetime         not null
#  account_id    :uuid             not null
#  artifact_id   :uuid             # <-- in associations section
#  user_id       :uuid
```

After this fix:
```ruby
#  id            :uuid             not null, primary key
#  amount_cents  :decimal(10, 6)
#  model         :string           not null
#  provider      :string           not null
#  usage         :jsonb            not null
#  created_at    :datetime         not null
#  updated_at    :datetime         not null
#  account_id    :uuid             not null
#  artifact_id   :uuid             # <-- grouped together
#  artifact_type :string           # <-- grouped together
#  user_id       :uuid
```

## Testing

- Added a test case to verify polymorphic associations are grouped together
- All existing tests continue to pass
- Code passes linting standards


garriguv:
@drwl yep, that makes sense.

I added an additional `grouped_polymorphic` option. When on, it'll group columns polymorphic associations columns together when `classified_sort` is on.
It has no impact with sort since the columns are already sorted alphabetically in that mode.

garriguv:
Cool, thank you for merging this!

drwl:

drwl:
Also fyi, `bin/rails g annotate_rb:update_config` should update your config to add the new `grouped_polymorphic` option to the project yml file.

Issue #233, opened 2025-07-02
When the --show-migration option is specified, the schema version of the primary database is also annotated for models referring to the secondary database.


OdenTakashi:
This issue is part of the work for issue https://github.com/drwl/annotaterb/issues/188.
> This issue is part of the work for issue https://github.com/drwl/annotaterb/issues/188.

PR #241, opened 2025-07-19
fix: --show-migration to use per-model database connections

OdenTakashi:
Thank you for your continued maintenance.

Previously, the annotation builder used a global migration version that was cached in options state, which caused incorrect version numbers in multi-database configurations where different models connect to different databases.

This changes:
- Removes global version caching from options state
- Adds get_migration_version_for_model() method that gets version from each model's specific database connection
- Only calculates version when include_version option is enabled
- Maintains backward compatibility with single database setups

fixes: #233


drwl:

PR #242, opened 2025-07-22
Add AnnotateRb::Runner.running? method

thewatts:
This is so that we can facilitate dynamic logic during the annotation process.

This is something that the historical `annotate_models` gem provided, via `Annotate::Migration.class_variable_get(:@@working))`

https://github.com/ctran/annotate_models/blob/5d01c4171990c4fe7b9b0977b05ce14a98209e53/lib/tasks/annotate_models_migrate.rake#L36-L41

Instead of having to reach into the Migration class, this instead stores some state on the `AnnotateRb::Runner` class so that we can ask it directly if it's in the middle of running.

### DIFF
--- a/lib/annotate_rb/runner.rb
+++ b/lib/annotate_rb/runner.rb
@@ -3,9 +3,23 @@
 module AnnotateRb
   class Runner
     class << self
+      attr_reader :runner
+
       def run(args)
-        new.run(args)
+        self.runner = new
+
+        runner.run(args)
+
+        self.runner = nil
       end
+
+      def running?
+        !!runner
+      end
+
+      private
+
+      attr_writer :runner
     end
 
     def run(args)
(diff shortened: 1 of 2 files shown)


drwl:
@thewatts I'm happy to add this in, but for my own curiosity, would you mind sharing how you use this in your workflow?

thewatts:
Hey @drwl, sure!

We're doing this, in our model file:

Our production schema doesn't entirely line up with our dev/test schemas.

```ruby
class Log < ApplicationRecord
  if Rails.env.test? || (defined?(Annotate) && defined?(Annotate::Migration) && Annotate::Migration.class_variable_get(:@@working))
    self.table_name = :logs
  else
    self.table_name = :all_logs
  end

  # ... the rest
end
```

This update would allow us to do:

```ruby
class Log < ApplicationRecord
  if Rails.env.test? || (defined?(AnnotateRb) && AnnotateRb::Runner.running?)
    self.table_name = :logs
  else
    self.table_name = :all_logs
  end

  # ... the rest
end
```

drwl (approved):
Thanks for adding this

drwl:
@thewatts thanks for the explanation!

PR #245, opened 2025-08-03
Run on rollback in app with multiple databases

### LINKED ISSUE #244: Does not run on rollback with multiple databases
Hello! I just switched an application over from the `annotate` gem. Thank you for keeping this gem alive.

---

When there are multiple databases, Rails _requires_ specifying the database in the rollback command, e.g. `rails db:rollback:primary` to rollback the primary database. Currently this gem does not account for the database as a suffix on rollback, only on migrate: https://github.com/drwl/annotaterb/blob/f8d22867eb0ce0d09ed3a92cedae00baac2eedf0/lib/annotate_rb/tasks/annotate_models_migrate.rake#L19-L22

z1lk:
Fixes #244.

I started out writing a spec for the Rake file, then realized that would depend on Rails being loaded, but an integration test felt awfully heavy and involved. I opted to update the Rake specs to support stubbing the Rails interfaces it relies on.


drwl (approved):
Thanks for submitting an issue and adding a fix! I don't have a multi db rails app to test this with, but the changes make sense to me

drwl:

Issue #254, opened 2025-08-08
Displaying database name for multi-database rails projects


professor:
Thank you for finishing the project: https://github.com/drwl/annotaterb/issues/188

I'm working on a rails project that has multiple databases and multiple schemas. While I was determining if it was safe to remove a column from a table, I thought the table was gone. While I was looking at the model's annotation, I missed that the table was in a different database.

Possible output:
```
# == Schema Information
#
# Table name: tasks
# Database name: second_database
#
#  id         :integer          not null, primary key
#  content    :string
#  count      :integer
#  status     :boolean
#  created_at :datetime         not null
#  updated_at :datetime         not null
#
class Task < SecondDatabaseRecord
```

If there is only one database, then the code would not print the database name.

OdenTakashi:
Hi @professor,

Thank you for providing the details. 
If @drwl is fine with this issue, I will take care of it on my end.

To ensure I'm on the right track, I have two quick questions about the behavior in this specific case:

1. In a multi-database environment, should models linked to the primary database not have the database name displayed, while only models linked to a secondary database do?

2. For a non-multi-database environment, is it correct that we never display the database name?

OdenTakashi:
> IMO, the database name should be listed regardless of if it is the primary or not. Not adding it would only cause confusion on why the database name isn't listed for certain models, until you dig into why.
Ya, I'd imagine the extra information isn't useful if you only have one databse.

I think the above policy seems reasonable,  so I will proceed based on this.

OdenTakashi:
I'm breaking the work into two separate PRs to
  keep the changes clean and easy to review.

  Here’s the plan:

  ---

  Step 1: Enhance the Test Environment

   * What: First, I'm going to update our test environment to support both single and
     multi-database configurations.
   * Why: The feature's requirement is twofold: the database name should be annotated
     in a multi-DB setup, but not in a single-DB setup. Our current test environment
     only allows us to validate the former. To properly test the complete behavior and
      prevent regressions, we need the ability to simulate a single-DB environment as
     well.
   * Note: This first PR will be strictly for testing infrastructure. It won't
     contain any feature logic.

  ---

  Step 2: Implement the Feature

   * What: Once the test environment is ready, I'll move on to the feature itself.
   * How: I'll add the logic to detect multi-DB environments and conditionally add
     the database name to the annotations. This PR will also include the new tests
     that validate both single and multi-DB scenarios.

  ---


  This two-step approach will make the process smooth and manageable. I'll be
[...]

PR #262, opened 2025-08-31
chore: introduce switchable DB environment for specs

OdenTakashi:
### Summary
This commit enhances the RSpec test environment to support both single and multi-database configurations.

### Problem
The next feature to be implemented is conditional model annotations, which requires asserting behavior in two different scenarios:

1. The annotation is added in a multi-DB environment.
2. The annotation is not added in a single-DB environment.

The existing test suite only supported a multi-DB setup, making it impossible to validate the second scenario.

### Solution
- The database.yml for the dummy app is updated to use an environment variable (SINGLE_DB_TEST). This allows it to conditionally exclude the secondary database configuration.
- A new RSpec shared_context ("when in a single database environment") is introduced. This provides specs with a clean, reusable way to enable the single-database mode for testing.

Refs #254

### DIFF
--- a/spec/dummyapp/config/database.yml
+++ b/spec/dummyapp/config/database.yml
@@ -11,10 +11,12 @@
     primary:
       <<: *default
       database: annotaterb_development
+    <% unless ENV['SINGLE_DB_TEST'] == 'true' %>
     secondary:
       <<: *default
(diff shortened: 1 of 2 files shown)


drwl (approved):
thanks for adding this!

PR #263, opened 2025-09-02
Escape column comment newlines when using "rightmost" value for position_of_column_comment option

rowanhogan:
This is a follow-up to the changes added in https://github.com/drwl/annotaterb/pull/199.

We noticed that when using the `rightmost` option for column comment position, any multi-line column comments were not being escaped in the same manner as the `with_name` option. This causes problems with generated model annotation as invalid uncommented newlines then appear. For example:

**Before**

```
#  notes                                        :text(55)         not null     Notes.
May include things like notes.
```

**After**

```
#  notes                                        :text(55)         not null     Notes\nMay include things like notes.
```

I have also updated the documentation for this option to include the available values – I had to dig through the codebase to find the possible values for the `--position-of-column-comments` option.

Thanks for this great gem, my company uses it in our projects 👏

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,7 +149,7 @@ Annotate model options:
         --without-comment            include database comments in model annotations
         --with-column-comments       include column comments in model annotations
         --without-column-comments    exclude column comments in model annotations
-        --position-of-column-comments VALUE
+        --position-of-column-comments [with_name|rightmost_column]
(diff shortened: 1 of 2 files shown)


drwl (approved):
Thanks for the bug fix and the kind note. Also thank you for improving test coverage 👍

PR #265, opened 2025-09-12
Honor --frozen option in routes

garriguv:
I found out that the `--frozen` does not work for `routes`, so I fixed it :)


drwl (approved):
Thanks for adding this along with tests. Historically, routes has been a neglected part of this gem so I am glad it's getting use and fixes 👍

PR #267, opened 2025-09-18
Make it possible to filter files by Regexp

matiaskorhonen:
Adds a new option that enables filtering model files by regexp.

Our use-case for this is filtering out spec files that live along side our models, for example we might have `user.rb` and `user.spec.rb` in the same directory.


OdenTakashi:
Thanks for opening this, @matiaskorhonen! Sorry for the late response. 

We're discussing a related topic around file filtering (allow/deny patterns and pattern syntax) over in #300 — feel free to chime in if you have thoughts !

PR #268, opened 2025-09-20
Update README.md

torgoton:
Fix typo - two options had the exact same description


drwl (approved):
Thank you for the catch

Issue #273, opened 2025-09-24
Option skip_on_db_migrate not honored


martinechtner:
Hi, the option `skip_on_db_migrate` does not seem to be honored when set to true in the yaml options file.

My config:

```ruby
# .annotaterb.yml

...
:skip_on_db_migrate: true
...

```

#### Expected behaviour

When running migrate commands annotations would not be run.

#### Actual behaviour

Annotations are run.

martinechtner:
Possible fix when calling `::AnnotateRb::Runner.run(["models"])` from https://github.com/drwl/annotaterb/blob/dd101d201e94366e9b9d0df213ebd68a8a67d2fd/lib/annotate_rb/tasks/annotate_models_migrate.rake#L32 a `task_type` could be passed in and then return early if the `task_type == "migrate" && @options[:skip_on_db_migrate]&.== true`.

PR #274, opened 2025-09-26
Honor skip_on_db_migrate config option when runnig migrate tasks

martinechtner:
Skip annotating on migration task when `skip_on_db_migrate` is set to `true`

Fixes https://github.com/drwl/annotaterb/issues/273


drwl (approved):
Looks good to me, thanks for ~adding~ fixing this

PR #281, opened 2025-10-21
Support .config directory convention

ngan:
In an effort to reduce the number of configuration dotfiles from the project root, I came across the `.config/` convention. RuboCop already supports this and it would be nice for this library to as well: https://docs.rubocop.org/rubocop/configuration.html#config-file-locations

This PR allows you to put your config files in the following location (ordered by precedence):
- `.annotaterb.yml` (default)
- `config/annotaterb.yml` (Rails friendly)
- `.config/.annotaterb.yml` (https://dot-config.github.io)
- `.config/annotaterb/config.yml` (https://dot-config.github.io)

More reading about this movement: https://dot-config.github.io

I've also taken the liberty of cleaning up somethings in this PR as well.

### DIFF
--- a/lib/annotate_rb/runner.rb
+++ b/lib/annotate_rb/runner.rb
@@ -23,18 +23,23 @@ module AnnotateRb
     end
 
     def run(args)
+      config_file_options = ConfigLoader.load_config
       parser = Parser.new(args, {})
+
       parsed_options = parser.parse
       remaining_args = parser.remaining_args
 
+      options = config_file_options.merge(parsed_options)
+
+      @options = Options.from(options, {working_args: remaining_args})
       AnnotateRb::RakeBootstrapper.call
 
-      options = ConfigLoader.load_config.merge(parsed_options)
-      @options = Options.from(ConfigLoader.load_config.merge(parsed_options), { working_args: remaining_args })
-
-      raise "Didn't specify a command" unless @options[:command]
-
-      @options[:command].call(@options)
+      if @options[:command]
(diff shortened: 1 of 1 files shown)


lib/annotate_rb/runner.rb
```diff
@@ -23,23 +23,18 @@ def running?
...
 
     def run(args)
-      config_file_options = ConfigLoader.load_config
       parser = Parser.new(args, {})
-
       parsed_options = parser.parse
       remaining_args = parser.remaining_args
 
-      options = config_file_options.merge(parsed_options)
+      AnnotateRb::RakeBootstrapper.call
```
drwl: It's been a while since I've touched this code, but what's the thinking around the reordering?

lib/annotate_rb/runner.rb
```diff
@@ -23,23 +23,18 @@ def running?
...
 
-      @options = Options.from(options, {working_args: remaining_args})
-      AnnotateRb::RakeBootstrapper.call(@options)
+      options = ConfigLoader.load_config.merge(parsed_options)
```
drwl: fyi `options` is not used here as complained by linter. 

Can run `bundle exec standardrb -A` to lint the rest

lib/annotate_rb/runner.rb
ngan: I needed to re-order to use `Rails.root`. But since I found out this gem is used for non-rails projects, I've restored the code.

PR #283, opened 2025-10-21
feat: Add "ignore_multi_database_name" option

t27duck:
New "modern" Rails apps default to using the "solid trifecta" of solidqueue, cache, and cable. The generator defaults to creating a separate database for each feature meaning all new Rails apps would be multi-database out of the box.

With the introduction of https://github.com/drwl/annotaterb/pull/272/ all Rails apps with the trifecta configured will have "Database: primary" appended to every annotated file even though there's really one database the developer "cares" about as the other gems provide their own models and handling.

This adds an "ignore_multi_database_name" configuration option which disables including the database name even if the app has multiple databases.

### DIFF
--- a/lib/annotate_rb/model_annotator/annotation/annotation_builder.rb
+++ b/lib/annotate_rb/model_annotator/annotation/annotation_builder.rb
@@ -74,6 +74,8 @@ module AnnotateRb
         private
 
         def multi_db_environment?
+          return false if @options[:ignore_multi_database_name]
+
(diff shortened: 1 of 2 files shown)


OdenTakashi:
@t27duck Thanks for adding this option.

@drwl This PR looks great to me. Everything seems to work as expected.

jonmcelroy-appfolio:
thanks for this, looking forward to the release!

PR #287, opened 2025-10-25
Load columns from schema to avoid AR ignored/only column configuration

### LINKED ISSUE #286: Fix missing annotation for Active Record ignored columns
Currently, columns that are ignored within Active Record models are not displayed as part of the schema annotation. e.g.

```
class Post < ApplicationRecord
  self.ignored_columns += ["foobar"]
end
```

... will result in the foobar column not appearing in the annotation comment. 

This behavior has caused problems when a developer has gone to remove an entry from the AR ignored_columns because they thought it was no longer needed because they were referencing the annotation comment where it did not appear. 

**Ideally,** I would like the column to exist in the annotation comment, because the column _is_ present in the database schema.

(I believe this feature is different than the annotaterb ignored-columns feature)

bensheldon:
Fixes #286

This PR recreates the Active Record logic for fetching the columns from the schema cache before they are filter based on Active Record's `ignored_columns`/`only_columns` configuration.

- `columns` is defined here: https://github.com/rails/rails/blob/5d4cfc3e73a77bd5ee4fba3ec42e8c3782c9ab02/activerecord/lib/active_record/model_schema.rb#L443-L450
- ...and populated here: https://github.com/rails/rails/blob/5d4cfc3e73a77bd5ee4fba3ec42e8c3782c9ab02/activerecord/lib/active_record/model_schema.rb#L609-L615

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,7 @@ Annotate model options:
         --without-table-comments     exclude table comments in model annotations
         --classes-default-to-s class Custom classes to be represented with `to_s`, may be used multiple times
         --nested-position            Place annotations directly above nested classes or modules instead of at the top of the file.
+        --load-columns-from-schema   Load columns from schema to avoid ActiveRecord ignored/only column configuration
 
 Annotate routes options:
     Usage: annotaterb routes [options]
--- a/spec/integration/load_columns_from_schema_spec.rb
+++ b/spec/integration/load_columns_from_schema_spec.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+require "integration_spec_helper"
+
+RSpec.describe "Load columns from schema with ignored columns", type: "aruba" do
+  let(:command_timeout_seconds) { 10 }
+  let(:model_file) { "app/models/test_default.rb" }
+
+  before do
+    copy_dummy_app_into_aruba_working_directory
+    reset_database
+    run_migrations
+  end
+
+  it "does not include ignored columns by default" do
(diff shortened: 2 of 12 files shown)


drwl:
Hi @bensheldon apologies on the months long delay in response -- the issue was helpful in providing context. I do _think_ that this could cause regressions for other users of this gem because adding this in its current state would mean ignored columns will appear in annotations.

Can this new behavior be gated behind an option flag with the default preserving the existing behavior?

README.md
```diff
@@ -155,6 +155,7 @@ Annotate model options:
         --without-table-comments     exclude table comments in model annotations
         --classes-default-to-s class Custom classes to be represented with `to_s`, may be used multiple times
         --nested-position            Place annotations directly above nested classes or modules instead of at the top of the file.
+        --load-columns-from-schema   Load columns from schema to avoid ActiveRecord ignored/only column configuration
```
drwl: This naming seems fine to me 👍

spec/integration/load_columns_from_schema_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:model_file) { "app/models/test_default.rb" }
+
+  before do
+    copy_dummy_app_into_aruba_working_directory
+    reset_database
+    run_migrations
+  end
+
+  it "does not include ignored columns by default" do
+    run_command_and_stop("bundle exec annotaterb models #{model_file} --force", fail_on_error: true, exit_timeout: command_timeout_seconds)
```
drwl: Does `--force` need to be here to trigger the changes?

PR #291, opened 2025-11-05
Add option to skip database name

alexrudall:
## Summary

Our CI checks were failing because the database environment in tests differs from development, causing annotation mismatches. The Database name: primary line would appear/disappear depending on the environment.

### The Github action affected

```
      - name: Check model annotations are up-to-date
        env:
          DATABASE_URL: postgres://postgres:password@localhost:5432/test
          RAILS_ENV: test
          RAILS_MASTER_KEY: ${{ secrets.RAILS_MASTER_KEY }}
        run: |
          bin/rails db:test:prepare
          bundle exec annotaterb models --force
          git diff --exit-code app/models || (echo "❌ Model annotations are not up-to-date. Please run 'bundle exec annotaterb models' locally and commit the changes." && exit 1)
```

## Changes

Added --skip-database-name flag to exclude database names from annotations in multi-database setups.

### Usage:
```
annotaterb models --skip-database-name
```

### Configuration:
```
skip_database_name: true
```

## Tests Added

- Parser spec verifying the CLI flag sets skip_database_name: true

### DIFF
--- a/spec/lib/annotate_rb/parser_spec.rb
+++ b/spec/lib/annotate_rb/parser_spec.rb
@@ -609,5 +609,14 @@ module AnnotateRb # rubocop:disable Metrics/ModuleLength
         expect(result).to include(nested_position: true)
       end
     end
+
+    describe "--skip-database-name" do
+      let(:option) { "--skip-database-name" }
+      let(:args) { [option] }
+
+      it "sets skip_database_name to true" do
+        expect(result).to include(skip_database_name: true)
+      end
+    end
   end
 end


OdenTakashi:
Hi @alexrudall
Thanks for the PR — I totally understand what you’re trying to solve here, and I really appreciate the work!

That said, since we already have the `ignore_multi_database_name` option, adding another flag with similar behavior (`--skip-database-name`) could make things a bit harder to maintain in the long run due to duplicated configuration.
Ideally, we could make the existing ignore_multi_database_name option configurable from the CLI instead.

What do you think?

alexrudall:
> ignore_multi_database_name

Ah, I missed this! Does it have the same functionality? If so that would solve my problem and this could be closed! :)

OdenTakashi:
Ah, I see — thanks for opening the PR!
Glad to hear it’s been resolved 😊

PR #292, opened 2025-11-13
Use `#lease_connection` if available

viralpraxis:
Checking out a connection via `#connection` should not be a problem here, but if this behavior is disallowed globally we should switch to `#lease_connection`.


viralpraxis:
Hey @drwl, could you take a look please?

viralpraxis:
Sorry for yet another ping @drwl, could you please take a look? 🙂

PR #296, opened 2025-11-19
Update Rails generator command in README

EstebanDalelR:
New rails 8+ uses generate


drwl (approved):
Thanks for adding this

Issue #289, opened 2025-10-31
Ability to whitelist only specific model patterns


mbkloster:
I am looking at rolling out this gem for models in a very large codebase (thousands of such models), and we have a number of tools in place, some locally and some in CI, that would choke at a large number of files being changed in a short time. Ideally, I’d like some way to set up an allowlist / whitelist such that we only include models that match a certain pattern, then roll the gem out more gradually.

One way that this could be accomplished out of the box would be to use a combination of `--model-dir` (to some dir without any actual models) and then `--additional-file-patterns`, but this seems rather hacky, and relies on me having a “non-functional” directory to point my model dir to.

Instead, I’d ideally like to have functionality that gives us the ability to set a specific allowlist / whitelist, then add file patterns to that, sort of like `--additional-file-patterns`, but acting with a “don’t include any models as a baseline” to start. This would give a lot more flexibility for teams that wish to (for whatever reason) limit the scope of this gem.

Is it possible this functionality could be included? Or perhaps is already included, but I'm missing a better way to do it. I’m happy to approach creating a PR myself if need be.

PR #300, opened 2025-12-02
feat: Ability to whitelist only specific model patterns (#289)

mbkloster:
Implements #289 

Adds a new parameter which, if given, restricts models annotated to match a given set of patterns.

I added both an integration spec and a (much simpler) unit test to this, but let me know if there are patterns or approaches you'd like to see tested separately. I was able to verify that this works on the sample project in this repo.

### DIFF
--- a/lib/annotate_rb/model_annotator/model_files_getter.rb
+++ b/lib/annotate_rb/model_annotator/model_files_getter.rb
@@ -16,10 +16,14 @@ module AnnotateRb
           model_dirs = options[:model_dir].flat_map { |model_dir| Dir[model_dir] }
           model_dirs.each do |dir|
             Dir.chdir(dir) do
+              only_files = options[:only_file_patterns]&.map { |pattern| Dir.glob(pattern) }&.flatten
               list = if options[:ignore_model_sub_dir]
-                Dir["*.rb"].map { |f| [dir, f] }
+                Dir["*.rb"]
+                  .select { |f| only_files.nil? || only_files.include?(f) }
+                  .map { |f| [dir, f] }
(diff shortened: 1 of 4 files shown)


drwl:
Sorry for the delay on this. I have looked at the PR a few times and it does make sense. I just need to take some time on how to best handle this, since it seems like there's demand for being able to filter models/files to annotate (allow list) as well as filter to ignore (deny list).

@OdenTakashi do you have any thoughts on how we might want to handle this?

OdenTakashi:
> it seems like there's demand for being able to filter models/files to annotate (allow list) as well as filter to ignore (deny list).

I agree. While I don't think we need to add a deny list option right now, I can see cases where both would be useful together.

Thanks for the contribution, @mbkloster! I'm in favor of adding this. Before merging, I'd like to consider a few points:

- **Option naming**: Since we'll likely add a deny list later, it would be good to choose names that make the pairing obvious from the start. For example, `--include-file-patterns` / `--exclude-file-patterns`, or `--allowlist-patterns` / `--denylist-patterns`. The current name `--only-file-patterns` doesn't have a natural counterpart and could feel inconsistent once the deny option is introduced.
- **Precedence rule**: We should decide upfront how allow and deny interact when both are specified. My suggestion: apply the allow list first (narrowing the set), then apply the deny list (removing from that set).
- **Structure**: The implementation should be designed so that adding a deny option later doesn't require a breaking change or major refactor.

Also, could you add the new option and a brief description to the README as part of this PR?

What do you think?

drwl:
Moved the content of this message to #369

PR #304, opened 2025-12-20
Include ignore_multi_database_name in generated config

OdenTakashi:
Thanks for always maintaining this project.

## Problem:
The ignore_multi_database_name option was implemented but not included in OTHER_OPTION_KEYS, so it was omitted from .annotaterb.yml files generated by `annotate_rb:install`.

## Solution:
Add ignore_multi_database_name to OTHER_OPTION_KEYS to ensure it is included in generated configuration files.


drwl (approved):
Thank you for adding this

PR #313, opened 2026-02-17
Added command example for applying annotations to README

zackerms:
# Motivation

The existing README includes a command example for skipping annotations, but the steps for applying
annotations are only described in prose, making them easy to overlook.

# Solution

Added a code block showing what happens when you run `bin/rails db:migrate` with AnnotateRb installed,
including sample log output.


OdenTakashi (approved):
lgtm!
Thank you for adding this.

Issue #309, opened 2026-02-05
Annotate postgres enums


andys:
Currently a Postgres enum column shows, eg:
```
#  state                    :enum             default("creating"), not null
```

The actual schema contains:

```
 t.enum "state", default: "creating", null: false, enum_type: "states"

create_enum :states, [
    "creating",
    "deploying",
    "completed",
    "failed",
    "superseded",
  ], force: :cascade
```

Which is more useful information for the developer.

drwl:
Should be implemented in 4.23.0

Issue #176, opened 2025-01-15
Add support for printing enum types


schpet:
given rails has [more support for postgres enums](https://blog.saeloun.com/2022/01/12/rails-7-adds-custom-enum-support-in-postgresql/) it would be nice if annotate included more info for a given enum, e.g. given i have a migration like this

```rb
    create_enum :billing_method, %w[agency_bill direct_bill_to_insured]
    add_column :policies, :billing_method, :enum, enum_type: :billing_method, null: true
```

we get an annotation like this:

```rb
# == Schema Information
#
# Table name: policies
#
#  id                             :uuid             not null, primary key
#  billing_method                 :enum
```

it would be nice to get more info on the enum, e.g.

```rb
# == Schema Information
#
# Table name: policies
#
#  id                             :uuid             not null, primary key
#  billing_method                 :enum             enum_type: billing_method
#
# Enums
#
#  billing_method  agency_bill, direct_bill_to_insured
```

drwl:
Hi @schpet thanks for submitting a detailed request. I think it shouldn't be too hard to add, but at the moment I do not have the time to work on it. 

However, I am adding labels to invite anyone to submit a PR for it. If you want to take a stab at adding the feature I would definitely be able to provide support there.

drwl:
Should be implemented in 4.23.0

PR #324, opened 2026-03-23
Support opt-in display of PostgreSQL enum types in annotations

OdenTakashi:
## Summary
This change adds an opt-in `show_enums` option for PostgreSQL model annotations.

When enabled, it:

- adds an `enum_type:` attribute to enum columns
- adds an **Enums** section below the schema
- lists only the enum types used by the table

This feature is supported only for PostgreSQL and is safely skipped for MySQL and SQLite.

`connection.enum_types` was introduced in Rails 7.0 ([rails/rails#41469](https://github.com/rails/rails/pull/41469)) and may return either a `String` or an `Array` depending on the Rails version ([rails/rails#54141](https://github.com/rails/rails/pull/54141)). This gem handles both formats.

| Return type | Rails versions |
| --- | --- |
| `String` (`string_agg`) | Rails 7.0 - 8.0.1 |
| `Array` (`array_agg`) | Rails 7.2.3+, 8.0.2+ |

## Example
Output follows the format proposed in #176. 

```rb
# == Schema Information
#
# Table name: orders
#
#  id      :bigint           not null, primary key
#  status  :enum             not null, enum_type: order_status
#  billing :enum             enum_type: billing_method
#
# Enums
#
#  billing_method  agency_bill, direct_bill_to_insured
#  order_status    pending, shipped, delivered
#
```

## Configuration
```rb
# .annotaterb.yml
show_enums: true
```

refs: #176 , #309


drwl (approved):
Thank you very much for adding this and sorry it took so long to review.

I know after this gets released we can close some issues 👍

OdenTakashi:
No worries at all — thank you for the review!

PR #327, opened 2026-04-21
Fix --show-migration option not working on Rails 7.2+.

OdenTakashi:
## Problem
`ConnectionAdapter#migration_context` was deprecated in Rails 7.1 and removed in Rails 7.2 (rails/rails#51162).
The existing code called `connection.migration_context.current_version`, which raises `NoMethodError` on Rails 7.2+.

## Solution
Use `connection.pool.migration_context` when available (Rails 7.1+), falling back to `connection.migration_context` for older versions.

### DIFF
--- a/lib/annotate_rb/model_annotator/model_wrapper.rb
+++ b/lib/annotate_rb/model_annotator/model_wrapper.rb
@@ -229,7 +229,14 @@ module AnnotateRb
 
         if @options.get_state(cache_key).nil?
           migration_version = begin
-            connection.migration_context.current_version
+            # Rails 7.1+ moved migration_context from ConnectionAdapter to ConnectionPool.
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
Very clear fix!

Also, good to have the version reference in place, so that we can drop unused code paths in the future.

OdenTakashi:
Thank you for the review and the kind words!

drwl (approved):
Thanks for fixing this. It's always a bit tricky as the Rails API changes across versions.

OdenTakashi:
Thanks for revewing this.

Since this kind of issue can break differently depending on the Rails version, I think it would be good to include Rails versions in the CI matrix as well. 

That should help us catch compatibility issues like this earlier, especially around deprecated or removed Rails APIs.

PR #328, opened 2026-04-23
Fall back to current task when no top-level task is set

jordan-brough:
When a task is invoked programmatically Rake's CLI parser never runs so `top_level_tasks` is empty.
Fall back to the current task name in that case.

Example:
```ruby
$ rails runner '
    require "rake"
    Rails.application.load_tasks
    Rake::Task["db:migrate"].invoke
  '
```

Results in:
```
.../ruby/3.4.8/lib/ruby/gems/3.4.0/gems/rake-13.4.2/lib/rake/task_manager.rb:59:in 'Rake::TaskManager#[]':
  Don't know how to build task '' (See the list of available tasks with `rake --tasks`)
  (RuntimeError)
```
without this fix.


OdenTakashi (approved):
Thanks for fixing this.

PR #331, opened 2026-05-05
Respect `--config-path` during config loading

OdenTakashi:
## Problem
#301 added support for loading a config file from a custom path via `--config-path`. However, #293 changed `Runner#run` so that the config is loaded before CLI arguments are parsed.

As a result, `--config-path` was only parsed after the config had already been loaded, meaning the custom path was not used when deciding which config file to read.

## Solution
Parse CLI arguments before loading the config in `Runner#run`.

This ensures that the value passed to `--config-path` is applied to `ConfigFinder.config_path` before `ConfigLoader.load_config` is called.

### DIFF
--- a/lib/annotate_rb/runner.rb
+++ b/lib/annotate_rb/runner.rb
@@ -31,13 +31,13 @@ module AnnotateRb
     end
 
     def run(args, config_file_options: nil)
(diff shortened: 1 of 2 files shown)


drwl (approved):
Thanks for adding the detailed write up, it was easy for me to understand.

PR #333, opened 2026-05-05
Preserve indentation when re-running annotaterb on nested-position annotations.

### LINKED ISSUE #332: Re-running annotaterb strips indentation from nested (indented) schema annotations

Given a namespaced model whose schema annotation is already indented (e.g. produced earlier by `--nested-position`, or hand-aligned to match the surrounding `module` / `class` indentation):

```ruby
# frozen_string_literal: true

module Blog
  class Post
    # == Schema Information
    #
    # Table name: blog_comments
    #
    #  id         :bigint           not null, primary key
    #  body       :text             not null
    #  post_id    :bigint           not null
    #
    # Represents a comment left on a blog post.
    class Comment < ApplicationRecord
    end
  end
end
```

Run `bundle exec annotaterb models` (no schema change required — see below).

## Expected behavior

- If the schema content has not changed, the file should be left untouched.
- If it has changed, the rewritten annotation should keep the same leading indentation as the existing block (4 spaces in the example).

## Actual behavior

The schema block is rewritten flush-left, even though the surrounding code (and the existing doc comment for the inner class) keeps its original indentation:

```ruby
# frozen_string_literal: true

module Blog
  class Post
# == Schema Information
#
# Table name: blog_comments
#
#  id         :bigint           not null, primary key
#  body       :text             not null
#  post_id    :bigint           not null
#
    # Represents a comment left on a blog post.
    class Comment < ApplicationRecord
    end
  end
end
```

This happens on every invocation, even when the schema has not changed at all — running `annotaterb models` repeatedly on the (already indented) file will strip the indentation again each time.

```yaml
# (the bug also reproduces with the default config; nested-position option is
# what produces the indented annotations in the first place, but the indentation
# loss happens regardless of whether --nested-position is passed on re-run)
```

## Environment

- AnnotateRb version: 4.22.0
- Ruby version: 4.0.3
- Ruby on Rails version: 8.1.3
- Database adapter (e.g., sqlite3, pg, mysql2): pg
- Database adapter version: 1.6.3

yamat47:
Closes #332.

When a schema annotation is placed inside a nested module/class (e.g. via `--nested-position`, or hand-aligned to match the surrounding indentation), re-running annotaterb stripped the indentation off every line of the block.

On top of that, the re-run happened even when the schema had not changed, because the diff generator never recognised an indented block as "the same" as the freshly generated, flush-left one.

This PR fixes both halves of that problem so that nested-position annotations are stable across re-runs.

### DIFF
--- a/lib/annotate_rb/model_annotator/annotated_file/updater.rb
+++ b/lib/annotate_rb/model_annotator/annotated_file/updater.rb
@@ -18,12 +18,25 @@ module AnnotateRb
         def update
           return "" if !@parsed_file.has_annotations?
 
-          new_annotation = wrapped_content(@new_annotations)
+          new_annotation = indent(wrapped_content(@new_annotations), existing_indentation)
 
           _content = @file_content.sub(@parsed_file.annotations) { new_annotation }
         end
 
         private
+
+        # Returns the leading whitespace of the existing annotation block so
+        # nested-position annotations keep their indentation across re-runs.
+        def existing_indentation
+          first_line = @parsed_file.annotations.lines.first
+          first_line&.match(/\A([ \t]*)/)&.[](1) || ""
+        end
+
+        def indent(content, indentation)
+          return content if indentation.empty?
+
+          content.lines.map { |line| (line == "\n") ? line : "#{indentation}#{line}" }.join
+        end
 
         def wrapped_content(content)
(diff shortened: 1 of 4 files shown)


OdenTakashi (approved):
Thanks for fixing this. Looks good to me.

PR #334, opened 2026-05-06
Fix annotation placement in nested files and around trailing user comments

yamat47:
## Problem

Two real-world layouts make annotaterb misplace or corrupt the schema annotation.

### 1. The end of an existing annotation gets mis-detected

`AnnotationFinder` walks backward looking for the first `#` separator. With a multi-paragraph user doc placed after the schema, the walk stops *inside* the user's doc:

```ruby
# == Schema Information
# Table name: users
# ...
#                                  # ← actual end of schema
# First paragraph of user doc.
#                                  # ← walk-back stops HERE (wrong)
# Second paragraph of user doc.
```

The first paragraph and the inner `#` are absorbed into the detected annotation range and silently lost when the annotation is rewritten.

The same logic fails when the annotation lacks its trailing `#` separator (e.g. manually trimmed): the walk stops at a section-internal `#`, and the trailing content row escapes the detected range — duplicating on the next run.

### 2. The annotation target line is wrong in nested files

`Generator#determine_annotation_position` previously used `parsed.starts.first` (no class case) and `parsed.starts.select { |name, _| type_map[name] == :class }.last` (class case). Both fail on common shapes:

```ruby
# spec file: parsed.starts.first picks `require`,
# so the annotation lands above the require line:
# == Schema Information       # ← wrong: should be just above RSpec.describe
# ...
require 'rails_helper'

RSpec.describe Sample do
end
```

```ruby
# self-reference inside the method body: `.last` picks
# the body reference, so the annotation lands inside the method:
class Sample < ApplicationRecord
  def call
    # == Schema Information    # ← wrong: should be above `class Sample`
    # ...
    Sample::Helper.run
  end
end
```

## Solution

Two small post-processors over `CustomParser`'s raw output.

**`walk_forward_through_schema`** (in `AnnotationFinder`, replaces the walk-back). Advances from the schema header while every successor matches `schema_like?` — a header prefix, an exact header (Indexes / Foreign Keys / Check Constraints), a tabular row (`#` + 2+ leading spaces), or a `#` separator. Stops at the last schema-like comment.

**`FileParser::AnnotationTarget.find`** (new). Skips `require` / `require_relative` / `load` and de-duplicates class entries by name so each class only contributes its declaration line. `Generator` and `ParsedFile` both go through it.

### DIFF
--- a/lib/annotate_rb/model_annotator/annotated_file/generator.rb
+++ b/lib/annotate_rb/model_annotator/annotated_file/generator.rb
@@ -59,17 +59,8 @@ module AnnotateRb
           content_with_annotations_written_before.join
         end
 
-        # Determines where to place the annotation based on the nested_position option.
-        # When nested_position is enabled, finds the most deeply nested class declaration
-        # to place annotations directly above nested classes instead of at the file top.
         def determine_annotation_position(parsed)
-          # Handle empty files where no classes/modules are found
-          return [nil, 0] if parsed.starts.empty?
-
-          return parsed.starts.first unless @options[:nested_position] && parsed.respond_to?(:type_map)
-
-          class_entries = parsed.starts.select { |name, _line| parsed.type_map[name] == :class }
-          class_entries.last || parsed.starts.first
+          FileParser::AnnotationTarget.find(parsed, @options) || [nil, 0]
         end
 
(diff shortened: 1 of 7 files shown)


drwl (approved):
Thanks for the write up and making these changes!

PR #336, opened 2026-05-09
Fix annotation placement above inner class declarations under `nested_position`

yamat47:
## Summary

When `nested_position` is enabled, `AnnotationTarget` (introduced in #334) picks the **deepest** class declaration as the insertion point. That heuristic breaks for any model that contains an inner class declaration — custom errors, value objects,
`Struct`/`Data` classes, etc. — because the deepest declaration is the inner one, so the schema annotation gets inserted directly above it instead of above the model class itself.

### Reproduction

```ruby
class Sample < ApplicationRecord
  include SomeMixin

  class ProcessingError < StandardError; end   # ← deepest, wrongly chosen

  def call
    raise ProcessingError
  end
end
```

Running annotaterb with nested_position lands the schema block right above class ProcessingError, not above class Sample.

## Fix

`ProjectAnnotator` already resolves each model file to its actual `ActiveRecord` class via `ModelClassGetter`, so we have the authoritative class name without needing any heuristic. This PR threads `klass.name.demodulize` through:

`SingleFileAnnotatorInstruction` → `SingleFileAnnotator`
→ `ParsedFile` / `AnnotatedFile::Generator`
→ `FileParser::AnnotationTarget.find`

as `model_class_name`:, and `AnnotationTarget.find` matches that name against `parser.type_map` first.

The deepest-declaration heuristic remains as a fallback for the cases where the name is absent or doesn't match anything in the source:

- related files (specs, factories, fixtures) don't get a `model_class_name`
- legacy file/class layouts where the declared class name doesn't match the resolved AR class name

So existing behavior is preserved everywhere except the inner-class case.

### DIFF
--- a/lib/annotate_rb/model_annotator/annotated_file/generator.rb
+++ b/lib/annotate_rb/model_annotator/annotated_file/generator.rb
@@ -5,7 +5,7 @@ module AnnotateRb
     module AnnotatedFile
       # Generates the file with fresh annotations
       class Generator
-        def initialize(file_content, new_annotations, annotation_position, parser_klass, parsed_file, options)
+        def initialize(file_content, new_annotations, annotation_position, parser_klass, parsed_file, options, model_class_name: nil)
           @annotation_position = annotation_position
           @options = options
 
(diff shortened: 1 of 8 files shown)


OdenTakashi (approved):
Thanks for fixing this. Looks good to me 🚀 
Your detailed explanation made it very easy to understand.

PR #337, opened 2026-05-11
  Add `position_in_class: after_doc` to keep class doc adjacent to class

### LINKED ISSUE #201: annotaterb is moving my class doc
The old gem would leave my class doc where I put it - right before the class:

```
# Schema information
# ...

# My doc about MyModel
class MyModel < ...
```

annotaterb is moving my class doc to the top of the file:

```
# My doc about MyModel
# Schema information
# ...
#
class MyModel < ...
```

I would very much like for annotaterb to leave my doc right before my class.  I don't see that as an option.

And as a footnote, I wonder what is the difference between top/before and bottom/after.

yamat47:
## Summary

Closes drwl/annotaterb#201.

Add a new `after_doc` value for `position_in_class` (and the other `position_in_*` options). When set, the schema annotation is inserted **above the contiguous documentation comment block** that sits directly above the class declaration, so the human-written doc stays adjacent to the class.

The default behavior (`before` / `top`) is unchanged.

## Problem

With the existing `before` / `top` positions, a documentation comment placed directly above the class declaration gets pushed above the schema annotation — away from the class it documents:

```ruby
# Source file:
# Doc about User
end

# After annotaterb today:
# Doc about User
# == Schema Information
# ...
end
```

 ## Fix

 Setting `position_in_class: after_doc` slots the annotation above the doc block, keeping the doc next to the class:

 ```ruby
 # With position_in_class: after_doc:
 # == Schema Information
 # ...
 # Doc about User
 end
 ```

 A "documentation comment" is defined as the contiguous comment block immediately above the class declaration with no blank line between. Recognized magic comments (`encoding`, `frozen_string_literal`, `shareable_constant_value`, `warn_indent`, `typed`, `rbs_inline`, plus Emacs/Vim modelines) are excluded so the annotation still slots between magic comments and the class doc:

 ```ruby
 # Source file:
 # frozen_string_literal: true
 # Doc about User
 end

 # With position_in_class: after_doc:
 # frozen_string_literal: true
 # == Schema Information
 # ...
 # Doc about User
 end
 ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,10 +191,10 @@ Additional options that work for annotating models and routes
     -f [bare|rdoc|yard|markdown],    Render Schema Information as plain/RDoc/Yard/Markdown
         --format
         --config_path [path]         Path to configuration file (by default, .annotaterb.yml in the root of the project)
-    -p [before|top|after|bottom],    Place the annotations at the top (before) or the bottom (after) of the model/test/fixture/factory/route/serializer file(s)
-        --position
-        --pc, --position-in-class [before|top|after|bottom]
-                                     Place the annotations at the top (before) or the bottom (after) of the model file
+    -p [before|top|after|bottom|after_doc],
+        --position                   Place the annotations at the top (before), bottom (after), or above the class documentation (after_doc) of the model/test/fixture/factory/route/serializer file(s)
+        --pc, --position-in-class [before|top|after|bottom|after_doc]
+                                     Place the annotations at the top (before), bottom (after), or above the class documentation (after_doc) of the model file
         --pf, --position-in-factory [before|top|after|bottom]
                                      Place the annotations at the top (before) or the bottom (after) of any factory files
         --px, --position-in-fixture [before|top|after|bottom]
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -191,10 +191,10 @@ Additional options that work for annotating models and routes
     -f [bare|rdoc|yard|markdown],    Render Schema Information as plain/RDoc/Yard/Markdown
         --format
         --config_path [path]         Path to configuration file (by default, .annotaterb.yml in the root of the project)
-    -p [before|top|after|bottom],    Place the annotations at the top (before) or the bottom (after) of the model/test/fixture/factory/route/serializer file(s)
-        --position
-        --pc, --position-in-class [before|top|after|bottom]
-                                     Place the annotations at the top (before) or the bottom (after) of the model file
+    -p [before|top|after|bottom|after_doc],
```
OdenTakashi: With the existing options, the names describe where the annotation appears when reading the file from top to bottom: `before` places it before the class block, and `after` places it after the class block.

Following that same logic, `after_doc` sounds like the annotation would appear after the doc block — in other words, between the doc and the class. But the actual behavior is the opposite.

I think `before_doc` feels more natural because it reads as “place the annotation before the doc block,” which matches what actually happens:

```ruby
# == Schema Information  ← before the doc
# Doc about User
end
```

OdenTakashi (commented):
Thanks for adding this!
One thing that still feels a little confusing to me is the name `after_doc`.

README.md
yamat47: @OdenTakashi 

Thanks!

I confused myself with the naming here. Since the annotation is placed before the documentation comment (consistent with how before/after describe top-to-bottom placement), before_doc makes much more sense. Fixed in e5bb069 — renamed across the option values, CLI help, README, and specs.

OdenTakashi: Thanks for the quick fix! before_doc is much clearer.

OdenTakashi (approved):
Looks good to me!

 This nicely avoids the issue where the documentation comment ends up above the annotation, away from the class it describes.

Issue #340, opened 2026-05-27
`annotaterb models --delete` doesn't remove annotations from files excluded via `exclude_*`


jeduardo824:
## Summary

When a related file type is excluded from annotation (e.g. `exclude_tests: true`), `annotaterb models --delete` does **not** remove annotations that were previously written to those files. The annotations become orphaned, and there's no built-in way to clean them up other than temporarily reverting the config.

## Steps to reproduce

1. With `exclude_tests: false` (the default), annotate — the model **and** its test files get an annotation block.
2. Decide you no longer want annotations in test files and set `exclude_tests: true`.
3. Run `annotaterb models --delete`.

**Expected:** the annotation blocks are removed from the test files as well.

**Actual:** the model/fixture annotations are removed, but the test files keep their (now stale) annotation blocks. The only workaround is to set `exclude_tests: false`, run `--delete`, then set it back to `true`.

The same applies to any related type that can be excluded (fixtures, factories, serializers, scaffolds, controllers, helpers).

## Root cause

`ProjectAnnotationRemover#build_instructions_for_file` builds its related-file list with the same `RelatedFilesListBuilder` that is used for annotating:

https://github.com/drwl/annotaterb/blob/9538099/lib/annotate_rb/model_annotator/project_annotation_remover.rb#L51

`RelatedFilesListBuilder#build` gates each related file type on the `exclude_*` options, so an excluded type is never added to the list — and removal therefore never visits it.

## Proposed fix

The `exclude_*` options describe where annotations should be *added*; they shouldn't block *removal*. When removing, the builder should reach every related file type regardless of the exclude settings (removing an annotation from a file that has none is a no-op).

I have a small patch + unit test ready and am happy to open a PR if this direction sounds good to you.

OdenTakashi:
Hi @jeduardo824,

Thanks for opening this — I understand the issue.

I’m in favor of fixing it. In other words, `--delete` should also remove annotations from excluded file types. That seems consistent with the README description of the `exclude` option: **do not annotate**

One concern is that this may be a breaking change for users who rely on the current behavior. For example:

```rb
annotaterb models --delete -e tests
```

This can be used to remove annotations from models/fixtures while keeping test files untouched.

That said, deletion is probably not the main workflow for this gem. Also, using `-e/--exclude` to scope deletion is not really the original intended use of that option.

So I’m also fine with deferring support for that workflow until it becomes a real pain point for users.

@drwl — what do you think is the best direction here?

[...]

PR #343, opened 2026-06-08
Remove annotations from related files regardless of exclude_* options

jeduardo824:
When a related file type is excluded (e.g. `exclude_tests: true`), `annotaterb models --delete` skipped those files, leaving previously written annotations orphaned with no built-in way to clean them up.

ProjectAnnotationRemover reuses RelatedFilesListBuilder, which gates each related type on the `exclude_*` options for both annotation and removal. Add a `for_removal:` flag (default false) that bypasses those guards so removal reaches every related file; the annotate path is unchanged. Removing an annotation from a file that has none is a no-op.

Refs #340

### DIFF
--- a/lib/annotate_rb/model_annotator/project_annotation_remover.rb
+++ b/lib/annotate_rb/model_annotator/project_annotation_remover.rb
@@ -48,7 +48,7 @@ module AnnotateRb
         model_instruction = SingleFileRemoveAnnotationInstruction.new(file, @options)
         instructions << model_instruction
 
-        related_files = RelatedFilesListBuilder.new(file, model_name, table_name, @options).build
+        related_files = RelatedFilesListBuilder.new(file, model_name, table_name, @options, for_removal: true).build
         related_file_instructions = related_files.map do |f, _position_key|
           _instruction = SingleFileRemoveAnnotationInstruction.new(f, @options)
         end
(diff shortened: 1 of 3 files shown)


OdenTakashi (approved):
Thanks for fixing this. Looks great to me!

However, as I mentioned in issue #340, there is a possibility that a certain usage pattern may no longer work after this change, so I’d like @drwl  to take a look at that point.

That said, I don’t think that usage pattern is actually intended, so I’m personally fine with merging this as-is.

drwl:
I do think it's an issue. Between this and the other issues, I'm thinking that an approach forward would be to add another option to 'delete all' annotations regardless of any inclusions/exclusions. Thoughts? @jeduardo824 @OdenTakashi

OdenTakashi:
@drwl @jeduardo824
I agree with @drwl that adding a separate option, such as `--delete-all`, would be better than changing the current behavior of `--delete`.

My earlier approval focused on fixing #340, but on reflection, making `--delete` ignore `exclude_*` would give those options a different meaning for deletion than for annotation. I think it would be clearer to keep their behavior consistent and add an explicit flag for deleting everything regardless of filters.

What do you think, @jeduardo824?

jeduardo824:
I agree with you guys! I have a lot of work to do this week, but I’ll find some time to implement this over the weekend :)

@drwl @OdenTakashi

PR #351, opened 2026-06-26
Add Rails version matrix to CI

OdenTakashi:
##   Summary
Extend integration CI to run against Rails 7.2, 8.0, and 8.1 across all database adapters. 
Add Rails 8–specific expected-output templates only where schema annotation formatting differs (mainly datetime defaults), while keeping existing templates as the default fallback.

## Problem
Integration tests were pinned to a single Rails version (~> 7.1.0), so regressions on newer Rails releases could go unnoticed. When running against Rails 8, some integration specs failed because Rails 8 changed how datetime default values appear in schema annotations (e.g. nanosecond precision and UTC suffix). Updating every template globally would add noise and break expectations for older Rails
versions.

## Solution
• Add a rails dimension to the integration CI  matrix and pass RAILS_VERSION through dummyapp  setup and test execution.
• Make the dummyapp Gemfile respect RAILS_VERSION,  and allow sqlite3 2.x when testing Rails 8.
• Teach SpecHelper::Aruba#model_template to prefer  spec/templates/rails8/<adapter>/ when running on  Rails 8 and an override exists; otherwise fall back to the existing spec/templates/<adapter>/templates.
• Add Rails 8 overrides only for the specs that compare datetime output (12 files total), avoiding unnecessary duplication of the full template set.

## Note
The spec/templates/rails8/ overrides exist solely because Rails 8 changed how datetime default values are rendered in schema comments. This comes from rails/rails#52371(https://github.com/rails/rails/pull/52371), which updated TimeWithZone#inspect to use ISO 8601-style formatting (e.g. 2023-07-04 12:34:56...) instead of the RFC 822-style format used on Rails 7.2 and earlier (e.g. Tue, 04 Jul 2023 12:34:56...).

As long as we keep Rails 7.2 in the CI matrix, we need separate golden files for the specs that assert datetime defaults. If we drop Rails 7.2 support later, every tested Rails version would share the same output format, and we could remove the rails8/ directory and the version-specific template lookup in model_template — the matrix CI would still work with a single set of templates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,7 @@ jobs:
       matrix:
         adapter: [ 'mysql2', 'pg', 'sqlite3' ]
         ruby: ['3.3', '3.4', '4.0']
+        rails: ['~> 7.2.0', '~> 8.0.0', '~> 8.1.0']
 
     steps:
       - name: Checkout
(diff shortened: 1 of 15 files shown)


drwl (approved):
Thank you for improving the test suite 👍

PR #353, opened 2026-06-27
Keep fixture annotations out of ERB blocks (#345)

### LINKED ISSUE #345: Fixture files that start with <% (erb) insert doc inside the erb block

My yaml fixture
```
<%
define some hash
loop_times do
%>
some:
    block: of stuff
<%
end
%>
etc...
```

annotaterb models --force

## Expected behavior

The schema information should NOT be placed in the erb block.

## Actual behavior

The schema information goes into the erb block.

```
<%
# == Schema Information
# ...
define some hash
loop_times do
%>
some:
    block: of stuff
<%
end
%>
...
```

```yaml
---
active_admin: false
additional_file_patterns: []
routes: false
models: true
position_in_routes: before_doc
position_in_class: before_doc
position_in_test: before_doc
position_in_fixture: before_doc
position_in_factory: before_doc
position_in_serializer: before_doc
show_foreign_keys: true
show_complete_foreign_keys: false
show_indexes: true
simple_indexes: false
model_dir:
  - app/models
root_dir:
  - ""
include_version: false
require: []
exclude_tests: false
exclude_fixtures: false
exclude_factories: false
exclude_serializers: false
exclude_scaffolds: true
exclude_controllers: true
exclude_helpers: true
exclude_sti_subclasses: false
ignore_model_sub_dir: false
ignore_columns:
ignore_routes:
ignore_unknown_models: false
hide_limit_column_types: integer,bigint,boolean
hide_default_column_types: json,jsonb,hstore
skip_on_db_migrate: false
format_bare: true
format_rdoc: false
format_yard: false
format_markdown: false
sort: false
force: false
frozen: <%= !ENV["TESTING_ANNOTATED_MODELS"].to_s.empty? %>
classified_sort: true
trace: false
wrapper_open:
wrapper_close:
with_comment: true
```

## Environment

- AnnotateRb version: github main branch
- Ruby version: 4.0
- Ruby on Rails version: 8.0
- Database adapter (e.g., sqlite3, pg, mysql2): mysql2 0.5.6

Halvanhelv:
## Summary

Fixes #345.

Schema annotations were being written **inside** ERB tags in dynamic fixture files. For example, given a fixture that starts with a multi-line ERB block:

```yaml
<%
  total = 2
%>
<% total.times do |i| %>
user_<%= i %>:
  name: User <%= i %>
<% end %>
```

running `annotaterb models` produced:

```yaml
<%
  total = 2
# == Schema Information
# ...
%>
<% total.times do |i| %>
...
```

i.e. the annotation landed in the middle of the ERB block, corrupting the file.

## Root cause

ERB fixtures cannot be parsed by Psych, so `YmlParser` evaluated the ERB (`ERB.new(input).result`) and read Psych's line numbers off the **evaluated** output. Those line numbers were then used as indices into the **original** file.

When an ERB tag spans multiple lines, the evaluated output has a different line count than the source, so the offsets no longer line up and the annotation is inserted at the wrong position — inside an ERB tag, or inside a `<% ... do %>` loop. Evaluating the ERB also executes arbitrary Ruby, which can fail outright.

This affected every ERB fixture, not just files starting with `<%` — the existing `<% 1.upto(100) do |i| %>` case also had its annotation written inside the loop.

## Fix

Stop evaluating the ERB. Instead derive the content bounds directly from the original lines: the first non-blank, non-comment line is the start, so annotations are written above the ERB body (and after any leading comments), never inside a tag. The result is also idempotent across repeated runs.

## Tests

- New `YmlParser` spec for a fixture that starts with a multi-line ERB block.
- New `Generator` specs covering an ERB fixture that starts with `<%` and one preceded by a comment.
- Updated the existing ERB specs in `YmlParser` and `SingleFileAnnotator` whose expectations encoded the buggy placement.

Verified these specs fail without the fix and pass with it. Full unit suite is green.

### DIFF
--- a/lib/annotate_rb/model_annotator/file_parser/yml_parser.rb
+++ b/lib/annotate_rb/model_annotator/file_parser/yml_parser.rb
@@ -43,10 +43,15 @@ module AnnotateRb
           begin
             parser.parse(@input)
           rescue Psych::SyntaxError => _e
-            # "Dynamic fixtures with ERB" exist in Rails, and will cause Psych.parser to error
-            # This is a hacky solution to get around this and still have it parse
-            erb_yml = ERB.new(@input).result
-            parser.parse(erb_yml)
+            # "Dynamic fixtures with ERB" exist in Rails and cause Psych.parser to error.
+            #
+            # We deliberately do not evaluate the ERB and read line numbers off the
+            # result: evaluating runs arbitrary code, and the line numbers from the
+            # evaluated output do not map back to the original file (ERB tags spanning
+            # multiple lines shift the offsets), which would place annotations inside
+            # an ERB tag. Instead we derive the content bounds straight from the
+            # original lines so annotations land around the ERB body.
+            return record_erb_positions
           end
 
           stream = parser.handler.root
@@ -67,6 +72,29 @@ module AnnotateRb
             @starts << [nil, stream.end_line]
             @ends << [nil, stream.end_line]
           end
+        end
+
+        # Locates the content bounds of an ERB fixture directly from the original
+        # lines, treating the ERB/YAML body as the doc. The start is the first
+        # non-blank, non-comment line so annotations are written above the ERB
+        # block (and after any leading comments), never inside a tag.
+        def record_erb_positions
+          lines = @input.split($/)
+          content_start = lines.index { |line| content_line?(line) }
+
+          if content_start.nil?
+            @starts << [nil, 0]
+            @ends << [nil, 0]
+          else
+            content_end = lines.rindex { |line| content_line?(line) }
+            @starts << [nil, content_start]
+            @ends << [nil, content_end + 1]
+          end
+        end
+
+        def content_line?(line)
+          stripped = line.strip
+          !stripped.empty? && !stripped.start_with?("#")
         end
       end
     end
--- a/lib/annotate_rb/model_annotator/file_parser/yml_parser.rb
+++ b/lib/annotate_rb/model_annotator/file_parser/yml_parser.rb
@@ -51,6 +51,11 @@ module AnnotateRb
             # multiple lines shift the offsets), which would place annotations inside
             # an ERB tag. Instead we derive the content bounds straight from the
             # original lines so annotations land around the ERB body.
+            #
+            # Only fall back for actual ERB fixtures. Genuinely malformed YAML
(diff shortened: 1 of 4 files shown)


lib/annotate_rb/model_annotator/file_parser/yml_parser.rb
```diff
@@ -43,10 +43,15 @@ def parse_yml
...
-            parser.parse(erb_yml)
+            # "Dynamic fixtures with ERB" exist in Rails and cause Psych.parser to error.
+            #
+            # We deliberately do not evaluate the ERB and read line numbers off the
+            # result: evaluating runs arbitrary code, and the line numbers from the
+            # evaluated output do not map back to the original file (ERB tags spanning
+            # multiple lines shift the offsets), which would place annotations inside
+            # an ERB tag. Instead we derive the content bounds straight from the
+            # original lines so annotations land around the ERB body.
+            return record_erb_positions
```
OdenTakashi, line 59: One concern: this now rescues any `Psych::SyntaxError` and falls back to `record_erb_positions`, so invalid YAML without ERB.

 e.g.

```yaml
user_one: [
```

would no longer raise and could still be annotated.

Personally, I think surfacing a broken file is friendlier. Ruby model files generally fail while resolving/loading the model, so I wonder if YAML should keep similar behavior unless the file actually looks like an ERB fixture.

What do you think about re-raising when there are no ERB tags?

Halvanhelv: @OdenTakashi I agree, I added the update

lib/annotate_rb/model_annotator/file_parser/yml_parser.rb
```diff
@@ -68,6 +78,35 @@ def parse_yml
...
+        def content_line?(line)
+          stripped = line.strip
+          !stripped.empty? && !stripped.start_with?("#")
+        end
+
+        # True when the input contains an ERB tag, i.e. it is a dynamic fixture
+        # rather than plain (possibly malformed) YAML.
+        def erb_fixture?
+          @input.match?(/<%.*?%>/m)
+        end
```
OdenTakashi, line 109: I also like the `erb_fixture?` helper here. Even though the condition is small, it makes the guard really easy to read.

OdenTakashi (approved):
Thanks for the quick fix. Looks great to me.

PR #356, opened 2026-07-13
Add Markdown annotation idempotency test

aouxwoux:
Adds a regression test for re-running AnnotateRb on an already annotated Markdown model. It verifies the behavior merged in #354 without duplicating that implementation.

Validation:
- `bundle exec rake spec:unit`
- `bundle exec standardrb spec/lib/annotate_rb/model_annotator/single_file_annotator_spec.rb`
- `git diff --check`

### DIFF
--- a/spec/lib/annotate_rb/model_annotator/single_file_annotator_spec.rb
+++ b/spec/lib/annotate_rb/model_annotator/single_file_annotator_spec.rb
@@ -97,6 +97,50 @@ RSpec.describe AnnotateRb::ModelAnnotator::SingleFileAnnotator do
       end
     end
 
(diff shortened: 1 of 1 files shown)


drwl:
Thanks for submitting this PR. It seems like https://github.com/drwl/annotaterb/pull/354 covers the same changes and also includes a more in depth write up, so I went and merged that one. 

I'd hate to completely dismiss a contribution, but I did notice that the `single_file_annotator_spec.rb` change could still be valuable. If you want to submit that I'd be happy to merge that in.

OdenTakashi (approved):
Thanks for adding this. Looks good to me.

PR #357, opened 2026-07-13
added :ignore_database_name option

HoneyryderChuck:
when enabled, it won't write the database name to the annotatino block

this option may be confusing in a sharded database scenario, where the development name just does not map to reality

### DIFF
--- a/lib/annotate_rb/model_annotator/annotation/schema_header.rb
+++ b/lib/annotate_rb/model_annotator/annotation/schema_header.rb
@@ -49,7 +49,7 @@ module AnnotateRb
           [
             Components::BlankCommentLine.new,
             TableName.new(name),
-            (DatabaseName.new(database_name) if database_name),
+            (DatabaseName.new(database_name) if show_database_name?),
             Components::BlankCommentLine.new
           ].compact
         end
@@ -66,6 +66,10 @@ module AnnotateRb
 
         def display_table_comments?
           @options[:with_comment] && @options[:with_table_comments]
+        end
+
+        def show_database_name?
+          database_name && !@options[:ignore_database_name]
         end
 
         def name
--- a/lib/annotate_rb/options.rb
+++ b/lib/annotate_rb/options.rb
@@ -42,6 +42,7 @@ module AnnotateRb
       format_yard: false, # ModelAnnotator
       frozen: false, # ModelAnnotator, but should be used by both
       grouped_polymorphic: false, # ModelAnnotator
+      ignore_database_name: false, # ModelAnnotator
       ignore_model_sub_dir: false, # ModelAnnotator
       ignore_unknown_models: false, # ModelAnnotator
       include_version: false, # ModelAnnotator
@@ -118,6 +119,7 @@ module AnnotateRb
       :format_yard,
       :frozen,
       :grouped_polymorphic,
+      :ignore_database_name,
       :ignore_model_sub_dir,
       :ignore_unknown_models,
       :include_version,
--- a/lib/annotate_rb/parser.rb
+++ b/lib/annotate_rb/parser.rb
@@ -222,6 +222,11 @@ module AnnotateRb
         @options[:ignore_unknown_models] = true
       end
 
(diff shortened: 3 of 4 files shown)


drwl (approved):
Thanks for the PR and the explanation 👍

OdenTakashi:
Thank you for the PR! Sorry for the delayed reply.

Is this for a multi-db environment? If so, there is already an `ignore_multi_database_name` option for this. ( #283 )
That said, it defaults to false. Originally there was no option to ignore the database name, so we kept the default as false to avoid changing the existing behavior. You'll need to set it to true for it to take effect.
If `ignore_multi_database_name` covers your use case, I'd like to remove this new option since it would be redundant.

( Also, having an option that isn't very visible / is easy to miss feels like something we should fix. )

HoneyryderChuck:
@OdenTakashi our setup is not for multi database, just sharded. I had tried `ignore_multi_database_name` support before, and it didn't work either for our case (Database Name is still there). 

I also think that it's quite common to name your database differently per environment ("${name}_production" for example), so having it pop up in annotations gets confusing, hence why I think this option is justified. We were previously using `annotate` gem before, and had to change to `annotaterb` in the process of upgrading to rails 8 (which `annotate` no longer supports), and this was the only gotcha we found (`annotate` does not write database names).

OdenTakashi:
@HoneyryderChuck  Thanks for the clarification — I overlooked that your setup is sharded rather than multi-database. Sorry about that.

I tried this locally with a sharded configuration and confirmed that the database name is included there as well. In my test, setting `ignore_multi_database_name: true` suppressed it. The logic that adds `database_name` only checks whether there are multiple database configurations for the current environment (`configs_for(...).size > 1`), so a sharded configuration is treated the same way as a multi-database configuration here.

This is also the only place where `database_name` is added to the annotation data. When `ignore_multi_database_name` is `true`, `multi_db_environment?` returns `false`, so `database_name` is not passed through:

```ruby
database_name = @model.database_name if multi_db_environment?

# ...

def multi_db_environment?
  return false if @options[:ignore_multi_database_name]

  if defined?(::Rails) && ::Rails.env
    ActiveRecord::Base.configurations.configs_for(env_name: ::Rails.env).size > 1
  else
    false
  end
end
```

That said, the name `ignore_multi_database_name` does not quite match the sharding use case, which feels a little confusing. This gave me a new perspective on the existing option, so I may take a look at consolidating this behavior under `ignore_database_name` myself.

Thanks again for the clarification — this was a helpful insight.

drwl:
@OdenTakashi Thank you for looking deeper into this. I should have looked at the existing options to see if there was any pre-existing behavior, so thanks for catching that.

HoneyryderChuck:
@OdenTakashi just a clarification that the development config ain't sharded, which is probably why the multi db option didn't catch it. And we annotate in development.

OdenTakashi:
@HoneyryderChuck 
Thanks for the clarification. 

I want to make sure I understand what happened correctly: in development, where the configuration uses neither multiple databases nor sharding, was the database name still included in the annotation even with `ignore_multi_database_name: true`?

I’m curious about what caused that, since `ignore_multi_database_name: true` should skip the condition that adds the database name to the annotation.

If you have a minimal reproduction, it would be helpful if you could share it.

HoneyryderChuck:
@OdenTakashi my bad, turns out the issue was in the config file on my side (`: ignore_multi_database_name` had 2 entries). I confirm `: ignore_multi_database_name` works. So yeah, I agree with your earlier statement that only the option name could be made clearer.

OdenTakashi:
@HoneyryderChuck Thanks for checking and confirming!

That makes sense — I’m glad to hear that `ignore_multi_database_name` works on your side.

Since `ignore_database_name` better reflects the actual behavior, including sharded configurations, I think it makes sense to use it as the clearer, preferred option going forward.

Since some users may already rely on `ignore_multi_database_name`, I’ll look into consolidating the two options while preserving backward compatibility.

PR #358, opened 2026-07-14
Respect DB column defaults over `attribute :foo, default: X` overrides

kamipo:
`Model#column_defaults` reflects attribute overrides declared with `attribute :foo, default: X`, so annotations previously showed the overridden Ruby value instead of the actual DB default. Derive defaults from `columns_hash` and typecast via the schema-level cast type so annotations mirror `db/schema.rb`.

### DIFF
--- a/lib/annotate_rb/model_annotator/model_wrapper.rb
+++ b/lib/annotate_rb/model_annotator/model_wrapper.rb
@@ -68,12 +68,19 @@ module AnnotateRb
           connection.table_comment(@klass.table_name).present?
       end
 
-      # Returns the DB schema defaults (not affected by `attribute :foo, default: X`).
+      # Returns column defaults for annotations.
+      #
+      # `Model#column_defaults` reflects `attribute :foo, default: X` overrides,
+      # which would incorrectly show the Ruby-side default in annotations
+      # instead of the DB schema default. To preserve model-level decorations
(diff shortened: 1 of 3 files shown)


OdenTakashi (approved):
Thanks for the fix!

I agree with this change — Schema Information should reflect the DB schema default, not Ruby-side `attribute` overrides.

@drwl 
One thing to keep in mind for release: this change will update existing annotations for models that override DB defaults with `attribute`. For example:

```ruby
# DB default: "hello world!"
attribute :message, :string, default: "overridden by the model"
```

Before:
```rb
#  message :string default("overridden by the model")
```

After:
```rb
#  message :string default("hello world!")
```

Columns without a database default are unaffected, and datetime defaults that are not overridden should keep their existing display format. 

This might be a useful note for the release notes so users understand why their existing annotations might change after upgrading.

toydestroyer:
Hello!

I wonder if this can be configurable?

In our case we have many integer-backed Rails' enums, e.g.
```ruby
# migration
t.integer :status, default: 0, null: false

# model
enum :status, { pending: 0, filled: 1, cancelled: 2 }

# annotation before
# status :integer          default("pending"), not null

# annotation after
# status :integer          default(0), not null
```

While `0` is indeed what stored in the table, developers in our team never refer to the integer values directly. I.e. we do something like `order.update(status: :cancelled)` instead of `order.update(status: 2)`.

PR #362, opened 2026-07-17
Fix NoMethodError when handling malformed annotations

OdenTakashi:
## Problem
`annotaterb models` crashed when a model file had a schema annotation header without a contiguous annotation body. For example:

```rb
# == Schema Info
end
```

AnnotationFinder raises MalformedAnnotation when it finds ` == Schema Info` / `## Schema Info` but the contiguous comment block is only one line. The rescue then crashed:

```sh
Unable to process app/models/user.rb: ...MalformedAnnotation
bundler: failed to load command: annotaterb (...)

.../single_file_annotator.rb:34:in `rescue in call':
  undefined method `[]' for nil:NilClass (NoMethodError)

    warn "\t" + e.backtrace.join("\n\t") if @options[:trace]
```

## Cause
SingleFileAnnotator and SingleFileAnnotationRemover are class methods, so `@options` is never set. The MalformedAnnotation rescue accessed `@options[:trace]` and raised `NoMethodError` after the intended warning.

--delete hid the same bug: `ProjectAnnotationRemover` rescues the secondary `NoMethodError`, so the process continued with a second "Unable to process" warning instead of aborting.

## Fix
Use the `options` argument at both call sites so malformed annotations are skipped cleanly.

### DIFF
--- a/lib/annotate_rb/model_annotator/single_file_annotation_remover.rb
+++ b/lib/annotate_rb/model_annotator/single_file_annotation_remover.rb
@@ -18,7 +18,7 @@ module AnnotateRb
             parsed_file = FileParser::ParsedFile.new(old_content, "", parser_klass, options).parse
           rescue FileParser::AnnotationFinder::MalformedAnnotation => e
             warn "Unable to process #{file_name}: #{e.message}"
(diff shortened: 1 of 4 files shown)


drwl (approved):

drwl:
Good catch!

PR #364, opened 2026-07-19
Annotate PostgreSQL unique and exclusion constraints

kamipo:
## Summary

Adds annotation support for two PostgreSQL constraint types that annotaterb has been overlooking: **unique constraints** and **exclusion constraints**. Also fixes a related leak where the indexes backing those constraints were showing up in the `Indexes` section.

## Motivation

In our application we lean on `DEFERRABLE INITIALLY DEFERRED` unique and exclusion constraints so we can swap values across rows within a transaction (e.g. reordering `position` on a list without hitting a mid-transaction uniqueness violation). This is a first-class PostgreSQL feature exposed by Rails 7.1+ as `add_unique_constraint ..., deferrable: :deferred` / `add_exclusion_constraint ..., deferrable: :deferred`, but none of it ever made it into annotaterb's output — the `deferrable` flag was dropped, and unique/exclusion constraints didn't have dedicated sections at all.

Worse, the underlying indexes were still enumerated: since PostgreSQL implements unique/exclusion constraints as indexes with the same name, `connection.indexes(table)` returns them, and annotaterb was rendering them as ordinary index rows. So a `deferrable` unique constraint appeared as a plain `UNIQUE` index with no hint of its actual semantics.

## Changes

1. **Two new annotation sections**, opt-in via config or CLI flags:
   - `show_unique_constraints` / `--show-unique-constraints`
   - `show_exclusion_constraints` / `--show-exclusion-constraints`

   Both mirror the existing check-constraint annotation pattern (default, markdown, yard, rdoc formats). The output includes column list / expression, `USING` method, `WHERE` predicate, and `DEFERRABLE INITIALLY IMMEDIATE|DEFERRED`.

2. **Filter constraint-backed indexes out of the `Indexes` section**, matching `ActiveRecord::SchemaDumper#indexes_in_create`: indexes whose name matches a unique or exclusion constraint name are excluded. Without this the same object would show up twice once you enable the new sections.

## Example output

```
#
# Indexes
#
#  index_reservations_on_room_id  (room_id)
#
# Unique Constraints
#
#  unique_position  (position) DEFERRABLE INITIALLY DEFERRED
#
# Exclusion Constraints
#
#  no_overlap       (room_id WITH =, during WITH &&) USING gist WHERE (canceled = false) DEFERRABLE INITIALLY DEFERRED
```

## Compatibility

- PostgreSQL-only feature; the accessors (`connection.unique_constraints`, `connection.exclusion_constraints`, `supports_*_constraints?`) exist in Rails 7.1+.
- Guarded with `respond_to?(:supports_*_constraints?)`, so on other adapters or older Rails versions the new sections silently do nothing and the index filter is a no-op — no risk of raising in projects that don't opt in.
- Both `show_unique_constraints` and `show_exclusion_constraints` default to `false`, so no existing annotations change unless the user opts in. The index-filtering change **does** apply unconditionally (matching Rails' schema dumper), so users of unique/exclusion constraints on Rails 7.1+ will see those rows disappear from `Indexes` — this is the desired behavior since it removes double reporting, and the info is preserved (and enriched) once the new sections are enabled.

### DIFF
--- a/lib/annotate_rb/model_annotator.rb
+++ b/lib/annotate_rb/model_annotator.rb
@@ -27,6 +27,8 @@ module AnnotateRb
     autoload :FileParser, "annotate_rb/model_annotator/file_parser"
     autoload :ZeitwerkClassGetter, "annotate_rb/model_annotator/zeitwerk_class_getter"
     autoload :CheckConstraintAnnotation, "annotate_rb/model_annotator/check_constraint_annotation"
+    autoload :UniqueConstraintAnnotation, "annotate_rb/model_annotator/unique_constraint_annotation"
+    autoload :ExclusionConstraintAnnotation, "annotate_rb/model_annotator/exclusion_constraint_annotation"
     autoload :EnumAnnotation, "annotate_rb/model_annotator/enum_annotation"
     autoload :FileToParserMapper, "annotate_rb/model_annotator/file_to_parser_mapper"
     autoload :Components, "annotate_rb/model_annotator/components"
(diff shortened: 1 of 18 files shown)


drwl:
Thanks for submitting this PR. It's a decent size changed so I will need to take time to sit down and review it. The write up is helpful in understanding the problem.

I will plan to do it tomorrow.

drwl (approved):
Looks good to me. I think we'll want to keep an eye out if any people report regressions and then we can direct them to this pr.

PR #365, opened 2026-07-19
Show DEFERRABLE INITIALLY on foreign keys

kamipo:
## Summary

Emit `DEFERRABLE INITIALLY IMMEDIATE|DEFERRED` on foreign key annotations, alongside the existing `ON DELETE` / `ON UPDATE` metadata. Related to #364.

## Motivation

Rails exposes deferrable foreign keys via `ForeignKeyDefinition#deferrable`, but annotaterb was silently dropping the flag — a foreign key created with `add_foreign_key ..., deferrable: :deferred` was indistinguishable from a plain one in the annotation.

## Compatibility

Rails 7.0's `extract_foreign_key_deferrable` returns `true` for the initially-immediate case (7.1+ normalized this to `:immediate`); mapped `true` → `IMMEDIATE` for stable output. Guarded with `respond_to?(:deferrable)` for older Rails / other adapters.


OdenTakashi (approved):
Thanks for adding this. Looks great to me.

PR #378, opened 2026-08-10
Fix ArgumentError sorting unnamed foreign keys when one is composite

### LINKED ISSUE #377: Foreign-key annotation crashes with ArgumentError when unnamed FKs mix with a composite FK
`ForeignKeyComponentBuilder#formatted_name` falls back to the raw `foreign_key.column` when the foreign key has no name:

```ruby
def formatted_name
  @formatted_name ||= if foreign_key.name.blank?
    foreign_key.column
  else
    ...
```

On SQLite every foreign key is unnamed (`fk.name` is `nil`), so this fallback is taken for all of them — and for a *composite* foreign key (Rails 7.1+), `foreign_key.column` is an **Array** while ordinary foreign keys yield a String. The sort in `ForeignKeyAnnotation::AnnotationBuilder#build`:

```ruby
foreign_key_components = fks.sort_by { |fk| [fk.formatted_name, fk.stringified_columns] }
```

then compares e.g. `["student_id", "student_id"]` against `[["delivery_on", "menu_id"], "[delivery_on, menu_id]"]` and raises:

```text
ArgumentError: comparison of Array with Array failed
lib/annotate_rb/model_annotator/foreign_key_annotation/annotation_builder.rb:26:in 'Enumerable#sort_by'
```

Because the rake hook runs after db tasks, the practical symptom in a Rails app is that **every `db:migrate` / `db:rollback` exits non-zero after the migration itself succeeds** — an easy failure to misattribute.

## Reproduction

- SQLite adapter (any adapter that reports unnamed foreign keys)
- A table with both an ordinary foreign key and a composite one, e.g. `add_foreign_key :checkout_school_meals, :edu_meal_options, column: [:delivery_on, :menu_id], primary_key: [:delivery_on, :menu_id]`
- Run model annotation (or any db task with the rake hook installed)

Reproduced identically on 4.23.0 and 4.24.0 (the builder code is the same in both).

## Suggested fix

Use the existing `stringified_columns` as the no-name fallback so the sort key (and the rendered name) is always a String — `"[delivery_on, menu_id]"` matches the style `ref_info` already renders for composite keys. As a side effect, the `max_size` padding calculation (`fks.map(&:formatted_name).map(&:size)`) becomes correct too, since it currently measures `Array#size` (2) instead of the rendered width.

Happy to send a PR for this.

6temes:
Fixes #377

## Problem

`ForeignKeyComponentBuilder#formatted_name` falls back to the raw `foreign_key.column` when a foreign key has no name. On adapters that report unnamed foreign keys (SQLite reports all of them that way), a composite foreign key yields an Array while ordinary ones yield a String, and `AnnotationBuilder#build`'s `sort_by { |fk| [fk.formatted_name, fk.stringified_columns] }` raises `ArgumentError: comparison of Array with Array failed`. In a Rails app this makes every `db:migrate`/`db:rollback` exit non-zero after the migration completes, since the rake hook runs afterwards.

## Fix

Use the existing `stringified_columns` as the no-name fallback, so the sort key and the rendered name are always Strings. For a composite key that renders as `[delivery_on, menu_id]`, matching the style `ref_info` already uses. This also makes the `max_size` padding measure the rendered width instead of `Array#size`.

## Verification

- New spec (unnamed simple + unnamed composite key on one model) reproduces the exact `ArgumentError` without the fix and passes with it.
- Full suite failure count is unchanged versus `main` in my environment (the same 26 pre-existing failures on both, unrelated to this change).


OdenTakashi:
Thanks for fixing this. Looks good to me.

OdenTakashi (approved):

PR #380, opened 2026-08-10
Fix enum-backed integer columns annotations

### LINKED ISSUE #373: 4.24.0: enum-backed integer columns annotate raw default(0) instead of default("label")
### Summary

Since 4.24.0, annotations for integer columns backing a Rails `enum` render the raw
integer default instead of the enum label. Regenerating annotations on an unchanged
schema produces a diff on every such column:

```diff
-#  signing_provider_service :integer          default("idnow"), not null
+#  signing_provider_service :integer          default(0), not null

-#  base_portfolio           :integer          default("classic"), not null
+#  base_portfolio           :integer          default(0), not null

-#  role                     :integer          default("user"), not null
+#  role                     :integer          default(0), not null
```

4.23.0 produced the label form. In our app this touched ~45 files (models, serializers
and factories) purely from the gem bump, which fails the CI job that runs
`rails db:migrate:reset` and asserts a clean working tree.

### Environment

- annotaterb 4.24.0 (4.23.0 is fine)
- Rails 8.1, Ruby 4.0
- PostgreSQL 17

### Cause

I believe this is unintended fallout from #358 ("Respect DB column defaults over
`attribute :foo, default: X` overrides"), which rewrote `ModelWrapper#column_defaults`:

```ruby
@column_defaults ||= @klass.column_defaults.each_with_object({}) do |(name, value), result|
  column = @klass.columns_hash[name]
  schema_value = schema_default_for(column)
  result[name] = (value == schema_value) ? value : schema_value
end
```

`Model#column_defaults` casts the DB default through the *attribute* type, so for an
enum column it yields the label (`"idnow"`). `schema_default_for` casts through the
*column* type, yielding `0`. They differ, so the branch treats it as an
attribute-level override and substitutes the raw schema value — but for enums the
difference is inherent to how `ActiveRecord::Enum` decorates the attribute type, not
an `attribute ... default:` override.

Declaring an enum is enough to trigger it:

```ruby
class AccountRequest < ApplicationRecord
  enum :signing_provider_service, {idnow: 0, fidentity: 1}, prefix: true
end
# 4.23.0: default("idnow")
# 4.24.0: default(0)
```

### Expected

Enum-backed columns keep rendering their symbolic default, since that is the
meaningful value for a reader — while `attribute :foo, default: X` overrides continue
to show the DB schema default as #358 intended.

A possible discriminator: skip the substitution when the attribute type is an
`ActiveRecord::Enum::EnumType` (or, more generally, when the attribute type is a
decorator around the column's cast type rather than an independent default), instead
of inferring an override from value inequality alone.

I'm happy to put up a PR if you agree with that direction.

dmke:
This PR contains two changes, feel free to omit the latter if not desired.

---

### Part 1: Detect attribute default overrides by raw value

Restores enum-backed defaults in annotations. Reverts to `default("idnow")` instead of the raw `default(0)` emitted since v4.24.0.

Fixes #373.

### Part 2: Add `enum_default_format` option

Make the default shown for enum-backed columns configurable, via `enum_default_format` or `--enum-default-format`:

    default("idnow")      # label (default)
    default(0)            # raw
    default(0: "idnow")   # both

`label` keeps the historical output.


georgebancila (approved):

dmke:
Pinging @drwl for review?

mjankowski:
@OdenTakashi I see some recent merges from you ... would be great to get a review of this PR and fix this regression from the last release?

drwl:
Changes look good to me. Given that this PR has input from other folks, I'm going to assume that this has been tested.

drwl (approved):

lib/annotate_rb/model_annotator/model_wrapper.rb
```diff
@@ -73,14 +73,17 @@ def has_table_comments?
...
+      # attributes whose raw default was replaced.
       def column_defaults
         @column_defaults ||= @klass.column_defaults.each_with_object({}) do |(name, value), result|
           column = @klass.columns_hash[name]
-          schema_value = schema_default_for(column)
-          result[name] = (value == schema_value) ? value : schema_value
+          result[name] = if attribute_default_overridden?(name, column)
+            schema_default_for(column)
+          else
+            enum_default(name, column, value)
```
OdenTakashi, line 85: enum_default is called for all non-overridden columns, even though it only handles enums.

Would it be clearer to check defined_enums at the call site instead?
```rb
result[name] = if attribute_default_overridden?(name, column)
  schema_default_for(column)
elsif @klass.defined_enums.key?(name)
  enum_default(name, column, value)
else
  value
end
```

Then the guard inside enum_default can be removed. Same behavior, but the flow feels a bit clearer.

OdenTakashi: Ah, looks like this was merged already 😄
No worries — this was just a small readability suggestion.

PR #382, opened 2026-08-24
Show index operator classes in annotations

kamipo:
PostgreSQL indexes can specify a per column operator class (`opclass:`), but the index annotation dropped it entirely, so an index like

```ruby
t.index ["name"], using: :gist, opclass: :gist_trgm_ops
```

was annotated as `(name) USING gist`, losing the information that distinguishes it from a plain gist index.

Operator classes are per column, so they are now rendered next to the column, in the same order that ActiveRecord generates the index SQL in (column, opclass, order):

```
#  index_organizations_on_name_trgm  (name gist_trgm_ops) USING gist
```

`IndexDefinition` condenses per column options into a single value when every column shares the same one, so `opclasses` (and `orders`) can be either a Hash keyed by column or a bare value. Both are handled now, which also fixes ordered indexes losing their order when every column shares it: `t.index [:a, :b], order: :desc` was annotated as `(a,b)` and is now annotated as `(a DESC,b DESC)`.

### DIFF
--- a/lib/annotate_rb/model_annotator/index_annotation/index_component.rb
+++ b/lib/annotate_rb/model_annotator/index_annotation/index_component.rb
@@ -136,12 +136,22 @@ module AnnotateRb
 
         def columns_info
           Array(index.columns).map do |col|
-            if index.try(:orders) && index.orders[col.to_s]
-              "#{col} #{index.orders[col.to_s].upcase}"
-            else
-              col.to_s.gsub("\r", '\r').gsub("\n", '\n')
-            end
+            column = col.to_s.gsub("\r", '\r').gsub("\n", '\n')
+            opclass = column_option(:opclasses, col)
+            order = column_option(:orders, col)
+
+            [column, opclass, order&.upcase].compact.join(" ")
           end
+        end
+
+        # ActiveRecord condenses per column index options (`opclasses`, `orders`)
+        # into a single value when every column shares the same one, so the
+        # option can either be a Hash keyed by column or a bare value.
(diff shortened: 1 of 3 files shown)


drwl (approved):
Thank you for adding this. Before I merge it in -- will this preserve existing behavior for indexes that don't specify the opclass?

lib/annotate_rb/model_annotator/index_annotation/index_component.rb
```diff
@@ -136,13 +136,23 @@ def to_markdown
...
-            if index.try(:orders) && index.orders[col.to_s]
-              "#{col} #{index.orders[col.to_s].upcase}"
-            else
-              col.to_s.gsub("\r", '\r').gsub("\n", '\n')
-            end
+            column = col.to_s.gsub("\r", '\r').gsub("\n", '\n')
+            opclass = column_option(:opclasses, col)
+            order = column_option(:orders, col)
+
+            [column, opclass, order&.upcase].compact.join(" ")
```
kamipo, line 143: > will this preserve existing behavior for indexes that don't specify the opclass?

Yes, indexes without an opclass render exactly as before.

`column_option` returns `nil` both when `opclasses` is the default empty hash and when the adapter's `IndexDefinition` doesn't respond to `opclasses` at all, so it's dropped by `compact` and the output is unchanged.

drwl: Thanks for confirming

PR #385, opened 2026-08-29
Update annotated model in README

koeda-ta:
Thank you for developing and maintaining this great gem.

I usually use `annotate`, but while reading the `AnnotateRb` documentation, I noticed a section that might be slightly confusing for beginners. I’ve made a small change to address this. Could you please take a look and let me know what you think?

## Changes

I changed the model file name in the README example as follows:

```diff
- # Annotated (1): app/models/task.rb
+ # Annotated (1): app/models/user.rb
```

## Reason

Since the environment variable `ANNOTATERB_SKIP_ON_DB_TASKS` is introduced immediately after `task.rb`, readers might assume that `task.rb` and `DB_TASKS` are related.

However, `task.rb` is the name of a model file, while `DB_TASKS` refers to database-related Rake tasks. The two are unrelated.

Additionally, the `User` model is already used as an example in the “Preserving class documentation comments” section of the README. Using `user.rb` here as well helps avoid confusion and keeps the examples consistent throughout the documentation.


OdenTakashi:
@koeda-ta 
Thanks for the suggestion!

I agree that having `task.rb` right next to `ANNOTATERB_SKIP_ON_DB_TASKS` can be a bit confusing, and I think using `User` would make the example clearer.

That said, I’d prefer not to change just this one line. The README currently uses `Task` as the main model in the annotation example near the top (`Table name: tasks` / `class Task`), so changing only this output to `user.rb` would make the examples inconsistent in a different way.

Would you mind updating the other `Task` examples in the README to use `User` as well, so that the documentation consistently uses the same model throughout?

koeda-ta:
@OdenTakashi 
Thanks for the review!

That makes sense — I've updated the annotation example at the top of the README to use `User` instead of `Task`,
so it now matches the `user.rb` output below.

While I was at it, I also replaced the sample columns (content / count / status) with ones that read more naturally for a `User` model, keeping the integer and boolean types covered:

```
# == Schema Information
#
# Table name: users
#
#  id            :integer          not null, primary key
#  name          :string
#  email         :string
#  sign_in_count :integer          default(0), not null
#  admin         :boolean          default(FALSE), not null
#  created_at    :datetime         not null
#  updated_at    :datetime         not null
#
```

If you'd rather keep the original column names and only rename the model, I'm happy to scale it back.

OdenTakashi (approved):
Thanks for the update — this is much easier to follow. Using `User` throughout, including sample columns that fit a `User` model, looks good to me.

PR #386, opened 2026-08-30
Fix multi-database annotation removal fixture targeting

OskarEichler:
Annotation removal always passes `klass.table_name` to related-file discovery. With multiple database classes sharing a table name, removing annotations from a secondary model can therefore target the primary database fixture. Match annotation behavior by supplying the table name only for the primary connection.

Verified with the full 874-example unit suite on Ruby 4.0.6 and 3.2.11, the Rails 8.1 SQLite integration suite, StandardRB, and a focused multi-database removal model.


OdenTakashi (approved):
Thanks for fixing this. Looks great to me.

PR #387, opened 2026-08-30
Use native prefix predicate in YAML parser

OskarEichler:
YmlParser uses the Active Support alias `starts_with?` for a plain String prefix check. Use the native `start_with?` so this parser method does not depend on a global Active Support core extension.

Verified with the full 874-example unit suite on Ruby 4.0.6 and 3.2.11, Rails integration, StandardRB, and a standalone parser model.


OdenTakashi (approved):
Thanks, looks good to me.

PR #388, opened 2026-08-30
Resolve globbed model directories before class lookup

OskarEichler:
`model_dir` accepts glob patterns, but fallback class lookup strips the literal pattern rather than its matched directory. Expand configured roots first and remove an exact path prefix, allowing globbed model directories to resolve the intended class.

Verified with the full 874-example unit suite on Ruby 4.0.6 and 3.2.11, Rails integration, StandardRB, and a focused globbed-directory model.


OdenTakashi (commented):
Thanks for the fix! 

Could you also add a regression spec for the case where `model_dir` contains a glob pattern?

The existing spec covers file discovery, but I’d also like to make sure that class lookup works correctly using the path relative to the expanded model directory.

OdenTakashi (approved):
Thanks again! I went ahead and added a regression spec to ensure class lookup works correctly when model_dir contains a glob pattern.

PR #391, opened 2026-09-03
[MEDIUM] Reject executable lines in generated annotations

OskarEichler:
## Summary

- Validate the complete model annotation block immediately before reading or rewriting a target file.
- Reject any physical line that is neither blank nor a Ruby comment.
- Keep all valid default, Markdown, RDoc, YARD, and YAML annotation output unchanged.

## Security impact

**MEDIUM urgency.** Several annotation components interpolate database schema metadata such as constraint expressions, names, index details, and enum values. If one of those values contains a newline, the following text can escape the leading `#` and be written as executable Ruby into a model or related source file. A later application boot or test run can execute that generated line.

Exploitation requires control of relevant database schema metadata (or direct control of the low-level annotation string) plus a developer or CI run of AnnotateRb. The gem is normally a development tool, which limits exposure, but the resulting impact is source-code execution in that environment.

## Reproduction

A check-constraint component with an expression ending in a newline plus `raise 'generated-code-marker'` currently generates:

```ruby
#  trusted_name  (price > 0)
raise 'generated-code-marker'
```

Passing that annotation to `SingleFileAnnotator` writes the uncommented line into the model. With this patch, the call raises `ArgumentError` before changing the file.

## Verification

- Focused unsafe annotation reproduction writes an uncommented marker on main and is rejected with the target file unchanged on this branch.
- Focused valid multiline annotation is still written.
- Ruby 4.0.6 unit suite: 874 examples, 0 failures, 2 existing pending.
- StandardRB: 175 files, no offenses.
- All 98 `lib` Ruby files compile.
- `git diff --check`.

## Limitations

This validates model and related-file annotations at their final write boundary. Route annotation generation uses a separate path and prefixes every route-map row itself; configuration files and the project Rakefile remain trusted local code.

## Breaking changes

Callers that intentionally pass non-comment content through the low-level model annotation API now receive `ArgumentError`. Generated annotation blocks are documented and expected to contain comments only.


OdenTakashi (commented):
Thanks for fixing this.

Could we add a spec to verify that a non-comment line in the annotation raises an error and leaves the file unchanged? The happy path is already covered.

OdenTakashi (approved):
Thanks again! I went ahead and added a spec for a non-comment line in the annotation to ensure it raises an error and leaves the file unchanged.

PR #393, opened 2026-09-13
Remove unused `routes` option

OdenTakashi:
The `routes` boolean is leftover from the original annotate gem, where a single `annotate` run chose models vs routes via this flag. annotaterb already uses the `annotaterb routes` command instead, and post-migrate route annotation is gated by `auto_annotate_routes_after_migrate`. No call site reads `options[:routes]`.


drwl (approved):
Thank for you tidying up unused code

